PR #2, opened 2020-01-17
Bump to 2.164.0

feisuzhu:


Dockerfile.ubuntu
```diff
@@ -43,9 +42,12 @@ RUN curl -fsSL https://download.docker.com/linux/ubuntu/gpg | apt-key add - \
...
 WORKDIR /actions-runner
 COPY install_actions.sh /actions-runner
+
+# Put version here to prevent cache invalidation
+ENV GH_RUNNER_VERSION="2.164.0"
 RUN chmod +x /actions-runner/install_actions.sh \
-  && /actions-runner/install_actions.sh \
-  rm /actions-runner/install_actions.sh
+  && /actions-runner/install_actions.sh ${GH_RUNNER_VERSION} \
+  && rm /actions-runner/install_actions.sh
```
myoung34, line 50: good catch, thanks

start-instances
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
myoung34, line 1: out of curiosity whats this function provide? it doesn't seem to be called anywhere

feisuzhu: It's a convenient script for starting multiple runners, so it will be run manually.

PR #13, opened 2020-04-07
Update examples to match WORKDIR to volume mounts

patst:
I'm running several runners on a raspberry pi and had some trouble getting a docker action running.

Its important to have the workspace directory aligned for the volume mount on the host, in the action runner and  the WORKDIR environment variable. This is already documented in the README but some of the examples differed a little bit.

I adjusted them to have everywhere something like `/tmp/github-runner-your-repo` as directory.


myoung34:
Nice Find @patst

PR #18, opened 2020-04-26
Fixed REPO_URL setting without ACCESS_TOKEN

EduMelo:
The env var _SHORT_URL wasn`t being setted if the ACCESS_TOKEN was null


myoung34 (changes requested):
You can remove line 25 with this change

myoung34:
Thanks!

PR #19, opened 2020-04-27
Remove runner on kill/shutdown

mvcaaa:
This is a topic for discussion

I am planning to use it in Kube cluster with auto-scaling.

I think, its better to remove agent from the pool when cluster kills an runner node/pod.  I cant see any negative feedback on this - there is no such thing as Runner history or something like that as in Teamcity eq (as far as i know). This will create only one extra API call.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -47,4 +47,11 @@ echo "Configuring"
     --unattended \
     --replace
 
+remove() {
+    ./config.sh remove --unattended --token "${RUNNER_TOKEN}"
+}
+
+trap 'remove; exit 130' INT
+trap 'remove; exit 143' TERM
+
 ./run.sh


myoung34:
This would be better done with something like dumb-init or tini imo

It wont help in all cases of hard aborts (instance goes away with no signalling etc) but it will help a little.

You could also use `--replace` and a cron pod that does smart cleanup for disconnected agents *in addition to* the remove() functionality

mvcaaa:
> It wont help in all cases of hard aborts (instance goes away with no signalling etc) but it will help a little.

agreed, didnt think about this

> You could also use --replace

replace is already here, but it wont help with dynamic runner hostnames

> cron pod that does smart cleanup for disconnected agents in addition to the remove() functionality

This is a nice idea, but ... this is another project :)

myoung34:
I'd open up an issue on https://github.com/actions/runner/issues/ and see if there's a server-side garbage clean

mvcaaa:
> I'd open up an issue on https://github.com/actions/runner/issues/ and see if there's a server-side garbage clean

There is no server released - only runner. Also - there is no gc afaik. And its not ideal. With docker caching/layering and stuffs i am running actions very-very fast even after `docker system prune`(how i test things atm) after each run

myoung34:
I mean github on their side might have a garbage cleaner or reaper process. It might be worth having an official answer

myoung34:
If you want to test this in dumb-init or tini id be happy to merge an entrypoint add-in there, until then I'm gonna close it and keep this in mind for other ideas if that's ok

Issue #27, opened 2020-06-03
Add note regarding RUNNER_WORKDIR volume mount


EmpireJones:
I spent the good part of a day trying to understand why github actions were behaving differently when using a local runner, compared to on github's runners. In particular, I was getting "file not found" errors when using actions, but when I did an `ls` on the files, via a workflow `run` command, the files were showing up just fine.

I believe what was happening is that because i was running the actions-runner in a docker container with the docker socket mounted from my host system, any docker actions would get a volume mount from the HOST system, rather than the runner's container.

The cause of things not working is that I was binding the RUNNER_WORKDIR to a different folder path on the host system, so the action runner container would see one thing, but the docker actions themselves would see another, because RUNNER_WORKDIR didn't exist on the host system.

It'd likely be good to add a note indicating that RUNNER_WORKDIR must exist on the host system and be the bind-mounted volume for RUNNER_WORKDIR in the container. Hopefully this saves others some time! 👍

myoung34:
Theres a precursor note about RUNNER_WORKDIR at https://github.com/myoung34/docker-github-actions-runner/blob/master/README.md

If you want to type up more information to help others in the situation, Id be happy to merge it in

myoung34:
Closed in #28 

Thanks again!

myoung34:
This will be out shortly in `latest-*` tags on docker hub

PR #28, opened 2020-06-03
Added RUNNER_WORKDIR mounting readme note

EmpireJones:
Fix for https://github.com/myoung34/docker-github-actions-runner/issues/27

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,9 @@ This has been tested and verified on:
  * armv7
  * arm64
  
-**NOTE: Only one runner can use the same RUNNER_WORKDIR if it is shared storage.**
+**NOTE:**
+- **Only one runner can use the same RUNNER_WORKDIR if it is shared storage.**
+- **RUNNER_WORKDIR must match the source path for the bind-mounted volume at RUNNER_WORKDIR, in order for container actions to access files.**
 
 ## Examples ##
 


README.md
```diff
@@ -11,7 +11,9 @@ This has been tested and verified on:
  * armv7
  * arm64
  
-**NOTE: Only one runner can use the same RUNNER_WORKDIR if it is shared storage.**
+**NOTE:**
+- **Only one runner can use the same RUNNER_WORKDIR if it is shared storage.**
+- **RUNNER_WORKDIR must match the source path for the bind-mounted volume at RUNNER_WORKDIR, in order for container actions to access files.**
```
myoung34, line 16: `- **RUNNER_WORKDIR must match the source path (and exist) for the bind-mounted volume at RUNNER_WORKDIR, in order for container actions to access files.**`

myoung34: i added `(and exist)` because that seemed like part of your finding also

myoung34 (changes requested):

README.md
EmpireJones: I think this shouldn't be a problem, per the docker docs:
> When the host directory of a bind-mounted volume doesn’t exist, Docker will automatically create this directory on the host for you.
- https://docs.docker.com/engine/reference/commandline/run/#mount-volume--v---read-only

What I meant by my original comment of "must exist on the host system" is that if the src and dest paths don't match, RUNNER_WORKDIR wouldn't exist on the host. If they do match though, it'll exist, per the docs mentioned above.

myoung34: OH. ok.

myoung34 (approved):

myoung34:
Thanks!

PR #30, opened 2020-06-12
Add support for clair scanning

myoung34:


myoung34:
Not worth the effort as older OS get more and more CVE

Almost none of these are outward facing anyway

PR #36, opened 2020-06-29
fix(docs): quote boolean in docker-compose example

aslafy-z:
Here's the relevant error when using the current snippet:

```
services.worker.environment.ORG_RUNNER contains true, which is an invalid type, it should be a string, number, or a null
```


myoung34:
Great catch!

I havent tested this outside of manual or nomad, so having these additions is super helpful

PR #37, opened 2020-06-29
feat: add unzip to docker image

aslafy-z:
Some actions (like https://github.com/hashicorp/setup-terraform) needs `unzip` which is a pretty tiny dependency that can be added seamlessly.


myoung34 (approved):

myoung34:
Nice! Thanks for this too.

Itll go up into the `*-latest` tags in the next few minutes

PR #55, opened 2020-09-18
Fix build removing apt-key adv docker

navarroaxel:
This step is not needed anymore https://github.com/docker/docker.github.io/issues/863

### DIFF
--- a/Dockerfile.base
+++ b/Dockerfile.base
@@ -39,7 +39,6 @@ RUN apt-get update && \
   && make \
   && make install \
   && cd / \
-  && apt-key adv --keyserver keyserver.ubuntu.com --recv-keys ${DOCKER_KEY} \
   && curl -fsSL https://download.docker.com/linux/ubuntu/gpg | apt-key add - \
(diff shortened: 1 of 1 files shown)


myoung34:
It's still needed because whatever they do in their installer isnt consistent across arm/amd64:

```
2020-09-18T13:01:57.5081398Z #7 287.1 W: GPG error: https://download.docker.com/linux/ubuntu xenial InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 7EA0A9C3F273FCD8
2020-09-18T13:01:57.5083515Z #7 287.1 W: The repository 'https://download.docker.com/linux/ubuntu xenial InRelease' is not signed.
```

which is what im fighting right now to get xenial/bionic/focal to be consistent across arm and amd64 😭

navarroaxel:
> arm/amd64

Oh, this is needed in focal `arm`?

myoung34:
It seems to be globally needed across the arm's for docker deb, so I suspect its a difference of packaging between amd64 and not-amd64

myoung34:
Everything seems to be finally working, and removing that definitely has a weird edge case, so im gonna close this unmerged.
Even if eventually its not needed I think its safe enough overall and an optimization I can revisit later

Thanks though! All the new images should be out in the next hour or so

PR #56, opened 2020-09-28
Add openssh client

pablosproject:
I find myself extending this docker image to add the ssh client, since I use [custom ssh repo key](https://github.com/webfactory/ssh-agent) and also use SFTP for some deploys.
I thnik the SSH client must be inside this base image.


myoung34 (approved):

myoung34:
LGTM

PR #57, opened 2020-10-05
Add support for the setup-python action

nberrington-cloudian:
The setup-python action [README.md](https://github.com/actions/setup-python/blob/main/README.md#linux) says self hosted runners need /opt/hostedtoolcache to exist and the AGENT_TOOLSDIRECTORY environment variable to point at it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,6 +2,9 @@
 FROM myoung34/github-runner-base:latest
 LABEL maintainer="myoung34@my.apsu.edu"
 
+ENV AGENT_TOOLSDIRECTORY=/opt/hostedtoolcache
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -2,6 +2,9 @@
 FROM myoung34/github-runner-base:latest
 LABEL maintainer="myoung34@my.apsu.edu"
 
+ENV AGENT_TOOLSDIRECTORY=/opt/hostedtoolcache
```
myoung34, line 5: env is only during build time, this wont exist during runtime

myoung34 (changes requested):

myoung34 (approved):

PR #58, opened 2020-10-09
Removed build from example docker-compose file

Jorgen1040:
All the build line in the example `docker-compose.yml` does is give an error when trying to run it so it would be better to remove it. (Unless there's something I don't know about)


myoung34:
I think this was a typo and it was supposed to be `context`

PR #62, opened 2020-10-23
Add the option to configure the RUNNER_GROUP on startup

damacus:
This adds the RUNNER_GROUP variable to configure which group the runner
is added to. Defaults to the "Default" runner group.


myoung34:
Thanks!

PR #64, opened 2020-11-02
Updates the base Dockerfile workflow to use Docker's GitHub Action

damacus:
We now use the newer docker setup & build actions.


myoung34:
Thanks! Ill make sure this keeps working post-merge and update it if not

PR #76, opened 2020-12-18
Update nomad example

Mongey:
There's an extra ` \`

I had to add `userns_mode = "host"` to get this to run, otherwise I'd get
```
failed to create container: API error (400): privileged mode is incompatible with user namespaces. You must run the container in the host namespace when running privileged mode
```


myoung34:
Thanks!

PR #78, opened 2020-12-25
Add locale in Dockerfile.base

nstdio:


Dockerfile.base
```diff
@@ -50,3 +53,4 @@ RUN apt-get update && \
   && chmod +x /usr/local/bin/docker-compose \
   && rm -rf /var/lib/apt/lists/* \
   && rm -rf /tmp/*
+  && locale-gen ${LANG}
```
myoung34, line 56: I think you also need to add `dpkg-reconfigure locales`

myoung34 (approved):

myoung34 (changes requested):
I think you also need to add `dpkg-reconfigure locales`

nstdio:
@myoung34 Can you please recheck?

Dockerfile.base
```diff
@@ -31,6 +31,10 @@ RUN apt-get update && \
     wget \
     dirmngr \
     openssh-client \
+    locales \
+  && dpkg-reconfigure locales \
```
myoung34: move this to after `locale-gen`

Dockerfile.base
```diff
@@ -31,6 +31,10 @@ RUN apt-get update && \
     wget \
     dirmngr \
     openssh-client \
+    locales \
+  && dpkg-reconfigure locales \ 
+  && locale-gen en_US.UTF-8 \
+  && export LANG=en_US.UTF-8 LANGUAGE=en_US:en LC_ALL=en_US.UTF-8 \
```
myoung34: these are lost on the running container, but i dont think theyre needed with the other 2 lines

myoung34 (changes requested):

Dockerfile.base
nstdio: Yeah. Just double checking we remove this line and use `-e LC_ALL=en_US.UTF-8` when running the container?

myoung34: I dont think so, but I'm unsure

myoung34 (changes requested):
if you add 
```
ENV LANG=en_US.UTF-8                                                                                                                                         
ENV LANGUAGE=en_US:en                                                                                                                                        
ENV LC_ALL=en_US.UTF-8 
```

to `Dockerfile.base` (after line 8: `ENV DOCKER_COMPOSE_VERSION="1.27.4"`) I can merge this, I tested it and it worked as expected. It will also allow overrides with `-e` which is ideal for future locales

myoung34 (approved):

myoung34:
Thanks!

PR #100, opened 2021-03-23
Dont install broken docker-compose on ARM

### LINKED ISSUE #99: docker-compose not available in arm image.
This line, which tries to install `docker-compose` using the Docker recommended process, fails on ARM:

https://github.com/myoung34/docker-github-actions-runner/blob/master/Dockerfile.base#L57

Instead, we get a file whose contents are `Not found`.

Docker does not actually release an ARM binary for Docker Compose. It seems to be a long standing problem with a lot of upvotes, and yet still no resolution:

https://github.com/docker/compose/issues/6831

The workaround that I am using on my host (where I have deployed the GitHub Actions runner) is the linuxserver image which does support ARM and is accessed via a `run.sh` shim:

https://hub.docker.com/r/linuxserver/docker-compose

This is working for me outside of the GitHub Actions runner, with one caveat. Environment variables from the shell are no longer automatically passed through to the compose container. Only special variables like `COMPOSE_FILE` and `COMPOSE_PROJECT_NAME` are automatically handled. For everything else, I have to set `COMPOSE_OPTIONS='-e FOO -e BAR -e BAZ'` to pass these through.

This is still a much better solution than the current completely broken `docker-compose` binary. I can probably just build my own runner image that has the `run.sh` shim installed as `/usr/local/bin/docker-compose`, but it'd be great to see it in the official image.

myoung34:
Resolves #99

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,3 +21,4 @@ COPY token.sh entrypoint.sh /
 RUN chmod +x /token.sh /entrypoint.sh
 
 ENTRYPOINT ["/entrypoint.sh"]
+CMD ["/actions-runner/bin/runsvc.sh"]
--- a/Dockerfile.base
+++ b/Dockerfile.base
@@ -36,6 +36,7 @@ RUN apt-get update && \
     openssh-client \
     locales \
     python3-pip \
+    jq \
   && pip3 install --no-cache-dir awscliv2 \
   && locale-gen en_US.UTF-8 \
   && dpkg-reconfigure locales \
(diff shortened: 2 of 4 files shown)


mrmachine:
It's good to have the limitation noted and a link to the workaround, thanks.

But it'd be better if we could install the `run.sh` shim instead, and note the caveat that users must export `COMPOSE_OPTIONS='-e FOO -e BAR` to pass through environment variables from their shell to compose?

myoung34:
If you'd like to add documentation to this after I merge about `COMPOSE_OPTIONS` I'd be happy to accept it, but I don't use docker-compose enough to be able to articulate that limitation.

I'll add the ability to use `run.sh` as a command override but I won't change it to that by default. I updated the PR to reflect the ability to override it with a docker command directive and default to `/actions-runner/bin/runsvc.sh` 

See the updated PR

myoung34:
its more related to scaling using docker-compose

If youd like to add a PR for a new flag that will cause a random work dir that would be acceptable on my side.

For the entrypoint: I wont alter the entrypoint outside of allowing a dynamic command though.

Your best bet is to set up automation that will do

```
FROM myoung34/github-runner:latest
COPY myentrypoint.sh /
RUN chmod +x /myentrypoint.sh
```

where `myentrypoint.sh` is something like

```
#!/usr/bin/dumb-init /bin/bash
if [[ ! -f /tmp/github-actions/runner_id.txt ]]; then
    head /dev/urandom | tr -dc A-Za-z0-9 | head -c 13 > /tmp/github-actions/runner_id.txt
fi
_RUNNER_ID="$(cat /tmp/github-actions/runner_id.txt)"
RUNNER_NAME="${RUNNER_NAME}-${_RUNNER_ID}"
RUNNER_WORKDIR="${RUNNER_WORKDIR}/${_RUNNER_ID}"

exec /entrypoint.sh run.sh
```

If you build that nightly it would work fairly seamlessly

PR #102, opened 2021-03-28
Sort apt packages and install dependencies for setup-ruby

rluisr:
- Sort packages by name
- Install some dependencies for setup-ruby compatible


myoung34:
What ruby version did you test with? I want to make sure the dependencies are as minimal as possible

rluisr:
I tested 2.4.4.

myoung34 (changes requested):
```
liblzma-dev
patch
g++-multilib
gcc-multilib
```

those additions shouldnt be necesary for ruby. sorting and adding `ruby-devel` is enough to add known apt dependencies

```
root@355bf77ed99f:/# ruby --version
ruby 2.7.0p0 (2019-12-25 revision 647ee6f091) [x86_64-linux-gnu]
```

Issue #116, opened 2021-06-19
Consider reusage of registered runner


NickVeld:
Your approach assuming reregistering a runner each a restart of its container. But why just not try to save the artifacts of the registration in case of the problems

I suppose I can implement such the possibility

myoung34:
I'm not against it but I'm not super interested in adding this myself

Runner registration takes only a few seconds, and runners are _generally_ meant to be long running (having none results in dropped actions)

I'm not sure I personally see the benefit, but I could see it being useful if it were added. I assume you'd just need to specify where the registration data lives so you can specify a volume (depending on which orchestrator is being used) but I'm not familiar with it.

uwehdaub:
We explicitly decided against reusing any container state, because it created some problems in our setup.
As a result we are now using a runner for only one step and then delete the old one and start a new one.
(Off course we had to adapt the entrypoint slightly.)

If this would be implemented, we need a possibility to opt-out.

NickVeld:
> 
> 
> We explicitly decided against reusing any container state, because it created some problems in our setup.
> As a result we are now using a runner for only one step and then delete the old one and start a new one.
> (Off course we had to adapt the entrypoint slightly.)
> 
> If this would be implemented, we need a possibility to opt-out.

Thanks. "Opt-out" is a quite simple: do not mount a directory and a state will stay inside.

myoung34:
resolved by #117

PR #117, opened 2021-06-27
Runner reusage

NickVeld:
Allows to avoid the runner reregistration and recreation after restarting.

So the Personal Access Token must be on the machine only for the first run of the runner.

Solves #116

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,7 @@ These containers are built via Github actions that [copy the dockerfile](https:/
 | `RUNNER_GROUP` | Name of the runner group to add this runner to (defaults to the default runner group) |
 | `GITHUB_HOST` | Optional URL of the Github Enterprise server e.g github.mycompany.com. Defaults to `github.com`. |
 | `DISABLE_AUTOMATIC_DEREGISTRATION` | Optional flag to disable signal catching for deregistration. Default is `false`. Any value other than exactly `false` is considered `true`. See [here](https://github.com/myoung34/docker-github-actions-runner/issues/94) |
+| `CONFIGURED_ACTIONS_RUNNER_FILES_DIR` | An optional path *inside a docker container* to mount the directory with the files of the registered runner. It allows avoiding reregistration each the start of the runner. Default is `/actions-runner-files` . |
 
 ## Examples ##
 
@@ -78,6 +79,22 @@ docker run -d --restart always --name github-runner \
   -e RUNNER_GROUP="my-group" \
   -v /var/run/docker.sock:/var/run/docker.sock \
   -v /tmp/github-runner-your-repo:/tmp/github-runner-your-repo \
+  myoung34/github-runner:latest
+```
+
+With the reusage of the registered runner assuming `CONFIGURED_ACTIONS_RUNNER_FILES_DIR=/actions-runner-files` (can be propogated to all other approaches):
+
+```shell
+# per repo
+docker run -d --restart always --name github-runner \
+  -e REPO_URL="https://github.com/myoung34/repo" \
+  -e RUNNER_NAME="foo-runner" \
+  -e RUNNER_TOKEN="footoken" \
+  -e RUNNER_WORKDIR="/tmp/github-runner-your-repo" \
+  -e RUNNER_GROUP="my-group" \
+  -v /var/run/docker.sock:/var/run/docker.sock \
+  -v /tmp/github-runner-your-repo:/tmp/github-runner-your-repo \
+  -v /path/on/host/system:/actions-runner-files
   myoung34/github-runner:latest
 ```
 
--- a/README.md
+++ b/README.md
@@ -82,10 +82,11 @@ docker run -d --restart always --name github-runner \
   myoung34/github-runner:latest
 ```
 
-With the reusage of the registered runner assuming `CONFIGURED_ACTIONS_RUNNER_FILES_DIR=/actions-runner-files` (can be propogated to all other approaches):
+Adding the reusage of the registered runner (can be propogated to all other approaches):
 
 ```shell
 # per repo
+export CONFIGURED_ACTIONS_RUNNER_FILES_DIR="/actions-runner-files"
 docker run -d --restart always --name github-runner \
   -e REPO_URL="https://github.com/myoung34/repo" \
   -e RUNNER_NAME="foo-runner" \
@@ -94,11 +95,11 @@ docker run -d --restart always --name github-runner \
   -e RUNNER_GROUP="my-group" \
   -v /var/run/docker.sock:/var/run/docker.sock \
   -v /tmp/github-runner-your-repo:/tmp/github-runner-your-repo \
-  -v /path/on/host/system:/actions-runner-files
+  -v /path/on/host/system:${CONFIGURED_ACTIONS_RUNNER_FILES_DIR}
   myoung34/github-runner:latest
 ```
 
-Or shell wrapper:
+Shell wrapper:
 
 ```shell
 function github-runner {
--- a/README.md
+++ b/README.md
@@ -86,16 +86,16 @@ Adding the reusage of the registered runner (can be propogated to all other appr
 
 ```shell
 # per repo
-export CONFIGURED_ACTIONS_RUNNER_FILES_DIR="/actions-runner-files"
 docker run -d --restart always --name github-runner \
   -e REPO_URL="https://github.com/myoung34/repo" \
   -e RUNNER_NAME="foo-runner" \
   -e RUNNER_TOKEN="footoken" \
   -e RUNNER_WORKDIR="/tmp/github-runner-your-repo" \
   -e RUNNER_GROUP="my-group" \
+  -e CONFIGURED_ACTIONS_RUNNER_FILES_DIR="/actions-runner-files"
   -v /var/run/docker.sock:/var/run/docker.sock \
   -v /tmp/github-runner-your-repo:/tmp/github-runner-your-repo \
-  -v /path/on/host/system:${CONFIGURED_ACTIONS_RUNNER_FILES_DIR}
+  -v /path/on/host/system:/actions-runner-files
   myoung34/github-runner:latest
 ```
 
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -14,7 +14,6 @@ deregister_runner() {
 run_runner() {
   # shellcheck disable=SC2068
   $@ 
-fi
 }
 
 _DISABLE_AUTOMATIC_DEREGISTRATION=${DISABLE_AUTOMATIC_DEREGISTRATION:-false}
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -52,14 +52,15 @@ case ${RUNNER_SCOPE} in
 esac
 
 # Loading the files from the mounted directory
-if [ -d ${CONFIGURED_ACTIONS_RUNNER_FILES_DIR} ]; then
-  cp -p -r ${CONFIGURED_ACTIONS_RUNNER_FILES_DIR}/. "/actions-runner"
+if [ -d "${CONFIGURED_ACTIONS_RUNNER_FILES_DIR}" ]; then
+  cp -p -r "${CONFIGURED_ACTIONS_RUNNER_FILES_DIR}/." "/actions-runner"
 fi
 
 if [ -f "/actions-runner/.runner" ]; then
   echo "The runner has already been configured"
   unset ACCESS_TOKEN
   unset RUNNER_TOKEN
+  # shellcheck disable=SC2068
   $@ 
   exit 0
 fi
@@ -83,8 +84,9 @@ echo "Configuring"
 unset RUNNER_TOKEN
 
 # Saving the files in another directory for the possibility to mount them from the host next time
-if [ -d ${CONFIGURED_ACTIONS_RUNNER_FILES_DIR} ]; then
-  cp -p -r "/actions-runner/_diag" "/actions-runner/svc.sh" /actions-runner/.[^.]* ${CONFIGURED_ACTIONS_RUNNER_FILES_DIR}
+if [ -d "${CONFIGURED_ACTIONS_RUNNER_FILES_DIR}" ]; then
+  # Quoting (even with double-quotes) the regexp brokes the copying
+  cp -p -r "/actions-runner/_diag" "/actions-runner/svc.sh" /actions-runner/.[^.]* "${CONFIGURED_ACTIONS_RUNNER_FILES_DIR}"
 fi
 
 if [[ ${_DISABLE_AUTOMATIC_DEREGISTRATION} == "false" ]]; then
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -11,10 +11,6 @@ deregister_runner() {
   exit
 }
 
-execute_docker_command() {
-  "$@" 
-}
-
 _DISABLE_AUTOMATIC_DEREGISTRATION=${DISABLE_AUTOMATIC_DEREGISTRATION:-false}
 
 _RUNNER_NAME=${RUNNER_NAME:-${RUNNER_NAME_PREFIX:-github-runner}-$(head /dev/urandom | tr -dc A-Za-z0-9 | head -c 13 ; echo '')}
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -81,6 +82,22 @@ docker run -d --restart always --name github-runner \
   myoung34/github-runner:latest
 ```
 
+With the reusage of the registered runner assuming `CONFIGURED_ACTIONS_RUNNER_FILES_DIR=/actions-runner-files` (can be propogated to all other approaches):
+
+```shell
+# per repo
+docker run -d --restart always --name github-runner \
+  -e REPO_URL="https://github.com/myoung34/repo" \
```
myoung34, line 90: i think you meant to add a call to `CONFIGURED_ACTIONS_RUNNER_FILES_DIR`  in this

entrypoint.sh
```diff
@@ -51,6 +51,19 @@ case ${RUNNER_SCOPE} in
     ;;
 esac
 
+# Loading the files from the mounted directory
+if [ -d ${CONFIGURED_ACTIONS_RUNNER_FILES_DIR} ]; then
```
myoung34: i dont think we want to check for the existence of the dir but rather the env var, this doesnt seem safe

entrypoint.sh
```diff
@@ -51,6 +51,19 @@ case ${RUNNER_SCOPE} in
...
+  cp -p -r ${CONFIGURED_ACTIONS_RUNNER_FILES_DIR}/. "/actions-runner"
+fi
+
+if [ -f "/actions-runner/.runner" ]; then
```
myoung34, line 64: this is new functionality, id rather copy the contents and behave as before, not change behavior from previous

running `$@` is already called at the last part of entrypoint.sh, and exit 0 is implicit already

entrypoint.sh
```diff
@@ -69,6 +82,11 @@ echo "Configuring"
 
 unset RUNNER_TOKEN
 
+# Saving the files in another directory for the possibility to mount them from the host next time
+if [ -d ${CONFIGURED_ACTIONS_RUNNER_FILES_DIR} ]; then
+  cp -p -r "/actions-runner/_diag" "/actions-runner/svc.sh" /actions-runner/.[^.]* ${CONFIGURED_ACTIONS_RUNNER_FILES_DIR}
```
myoung34: this doesnt seem inherently easy to understand, im not a fan of regex bash calls. it also seems different than line 56

myoung34 (changes requested):

README.md
NickVeld: Oops, strange. Thanks!

entrypoint.sh
NickVeld: I do not really know the threat from the checking the envvar for the directory existence, especially in the docker container. Do you suggest using the "do and ignore errors" approach?

myoung34: i think  checking for both the env var and the directory is fine, but only the directory with a default value can add a weird edge case. if its not set then i think you can short circuit and do nothing regardless of the directory presence

entrypoint.sh
NickVeld: 56 line just copy all files because it is is not really important what it copies. If the user will try to mess the directory it is his problem and you just cannot remove all such the possibilities.

87 line copies only the files that appears only after the registration:
* `svc.sh`
* `_diag`
* `.credentials`
* `.credentials_rsaparams`
* `.env`
* `.path`
* `.runner`

Considering the possible changes I thought about simplifying and not limiting to the certain hidden files but all

For the copying only hidden files see [this](https://superuser.com/questions/830562/how-to-copy-hidden-starting-with-a-dot-files-and-subdirectories-in-linux)

entrypoint.sh
NickVeld: I do not really get your suggestion but placing `$@; exit 0` is meant as a return statement for avoiding the runner reconfiguration

README.md
NickVeld: See "assuming `CONFIGURED_ACTIONS_RUNNER_FILES_DIR=/actions-runner-files`"
and line 97
`   -v /path/on/host/system:/actions-runner-files`

entrypoint.sh
NickVeld: How about the following: `if [ -d ${CONFIGURED_ACTIONS_RUNNER_FILES_DIR:=/actions-runner-files} ]; then` ?
https://stackoverflow.com/a/2013589

entrypoint.sh
myoung34: I mean we alreaady have a $@ and an exit 0

We should not have multiple run/exits but let it trickle through to the end

myoung34 (changes requested):

entrypoint.sh
myoung34: In this usage its a bit hard to read and a bit smelly to check on something that mutates (`:=`)

NickVeld: If not `:=` then for every such the check `:-` is needed

entrypoint.sh
NickVeld: Then only the way to put all the configuration code under `if` statement or in a function

README.md
NickVeld: Ok

entrypoint.sh
NickVeld: Why are you ignoring SC2068 instead of  using`"$@"`?

myoung34: I dont remember why SC2068, i think there was an interpolation error with docker run with that

README.md
```diff
@@ -81,7 +82,24 @@ docker run -d --restart always --name github-runner \
...
+export CONFIGURED_ACTIONS_RUNNER_FILES_DIR="/actions-runner-files"
+docker run -d --restart always --name github-runner \
+  -e REPO_URL="https://github.com/myoung34/repo" \
+  -e RUNNER_NAME="foo-runner" \
+  -e RUNNER_TOKEN="footoken" \
+  -e RUNNER_WORKDIR="/tmp/github-runner-your-repo" \
+  -e RUNNER_GROUP="my-group" \
+  -v /var/run/docker.sock:/var/run/docker.sock \
+  -v /tmp/github-runner-your-repo:/tmp/github-runner-your-repo \
+  -v /path/on/host/system:${CONFIGURED_ACTIONS_RUNNER_FILES_DIR}
```
myoung34: i'd remove this and do `-e CONFIGURED_ACTIONS_RUNNER_FILES_DIR=/tempdir`
and then `-v /path/on/host/system:/tempdir`

entrypoint.sh
```diff
@@ -11,6 +11,11 @@ deregister_runner() {
   exit
 }
 
+run_runner() {
```
myoung34: this isnt used anywhere it seems

entrypoint.sh
```diff
@@ -51,6 +56,24 @@ case ${RUNNER_SCOPE} in
...
+# Loading the files from the mounted directory
+if [ -d "${CONFIGURED_ACTIONS_RUNNER_FILES_DIR}" ]; then
+  cp -p -r "${CONFIGURED_ACTIONS_RUNNER_FILES_DIR}/." "/actions-runner"
+fi
+
+if [ -f "/actions-runner/.runner" ]; then
+  echo "The runner has already been configured"
+  unset ACCESS_TOKEN
+  unset RUNNER_TOKEN
+  execute_docker_command $@
```
myoung34: im not sure what `execute_docker_command` is

entrypoint.sh
```diff
@@ -69,9 +92,14 @@ echo "Configuring"
...
+  cp -p -r "/actions-runner/_diag" "/actions-runner/svc.sh" /actions-runner/.[^.]* "${CONFIGURED_ACTIONS_RUNNER_FILES_DIR}"
+fi
+
 if [[ ${_DISABLE_AUTOMATIC_DEREGISTRATION} == "false" ]]; then
   trap deregister_runner SIGINT SIGQUIT SIGTERM
 fi
 
-# shellcheck disable=SC2068
-$@ 
+execute_docker_command $@
```
myoung34: if the function just runs `$@` then its not needed, just keep `$@` here

myoung34: remove $@ and exit 0 from line 73-74, and just wrap the rest of entrypoint.sh in an `else`, that will let you have a single $@ and naturally exit

myoung34 (changes requested):

entrypoint.sh
NickVeld: This warning is about the following difference:

## Case 1:

You have `script.sh` with the following content:
```
exec $@
```

Then `sh script.sh "a" "b c"` cause the execution of the following command: `exec a b c`

## Case 2:

You have `script.sh` with the following content:
```
exec "$@"
```

Then `sh script.sh "a" "b c"` cause the execution of the following command: `exec "a" "b c"`

So SC2068 warn that you are falling into the first case. not the second

NickVeld: So `$@` can be changed to `"$@"` and there is no need to ignore SC2068

entrypoint.sh
NickVeld: It was changed

README.md
NickVeld: Ok

entrypoint.sh
NickVeld: Must be `execute_docker_command ()`

entrypoint.sh
NickVeld: See the `run_runner` thread

entrypoint.sh
NickVeld: I was frequently taught about removing the block indentation where it is possible because

1. Indents eat space for the symbols in the line (in case of the limit of the line length)
2. They are needed to be considered always in some specific cases

So there is my try to reduce the indentation. But if you are thinking that it must be removed so be it

The function is also increases readability (named action instead of just `$@`)

myoung34: I think its fine to do an if else and keep it indented. I also agree there but my preference for one $@ and one exit 0 is higher on my list

i think it should make it much easier to read in honesty

NickVeld: Done

entrypoint.sh
```diff
@@ -51,27 +51,51 @@ case ${RUNNER_SCOPE} in
...
-    --runnergroup "${_RUNNER_GROUP}" \
-    --unattended \
-    --replace
+# Loading the files from the mounted directory
+if [ -d "${CONFIGURED_ACTIONS_RUNNER_FILES_DIR}" ]; then
+  cp -p -r "${CONFIGURED_ACTIONS_RUNNER_FILES_DIR}/." "/actions-runner"
+fi
 
-unset RUNNER_TOKEN
+if [ -f "/actions-runner/.runner" ]; then
```
myoung34, line 64: This will mean that it always retains state and its no longer optional.

Others aside from myself want this to be an opt in functionality

entrypoint.sh
```diff
@@ -51,27 +51,51 @@ case ${RUNNER_SCOPE} in
...
+        --unattended \
+        --replace
+
+    # Saving the files in another directory for the possibility to mount them from the host next time
+    if [ -d "${CONFIGURED_ACTIONS_RUNNER_FILES_DIR}" ]; then
+      # Quoting (even with double-quotes) the regexp brokes the copying
+      cp -p -r "/actions-runner/_diag" "/actions-runner/svc.sh" /actions-runner/.[^.]* "${CONFIGURED_ACTIONS_RUNNER_FILES_DIR}"
+    fi
+
+    if [[ ${_DISABLE_AUTOMATIC_DEREGISTRATION} == "false" ]]; then
```
myoung34, line 91: we want to keep `_DISABLE_AUTOMATIC_DEREGISTRATION` even if they are retaining state

entrypoint.sh
```diff
@@ -51,27 +51,51 @@ case ${RUNNER_SCOPE} in
     ;;
 esac
 
-if [[ -n "${ACCESS_TOKEN}" ]]; then
-  _TOKEN=$(bash /token.sh)
-  RUNNER_TOKEN=$(echo "${_TOKEN}" | jq -r .token)
+# If the variable is not set, set it with the default value
+if [ -z "${CONFIGURED_ACTIONS_RUNNER_FILES_DIR}" ]; then
```
myoung34, line 55: this also makes it work by default vs opt-in

entrypoint.sh
```diff
@@ -51,27 +51,51 @@ case ${RUNNER_SCOPE} in
...
+    if [ -d "${CONFIGURED_ACTIONS_RUNNER_FILES_DIR}" ]; then
+      # Quoting (even with double-quotes) the regexp brokes the copying
+      cp -p -r "/actions-runner/_diag" "/actions-runner/svc.sh" /actions-runner/.[^.]* "${CONFIGURED_ACTIONS_RUNNER_FILES_DIR}"
+    fi
+
+    if [[ ${_DISABLE_AUTOMATIC_DEREGISTRATION} == "false" ]]; then
+      trap deregister_runner SIGINT SIGQUIT SIGTERM
+    fi
 
-if [[ ${_DISABLE_AUTOMATIC_DEREGISTRATION} == "false" ]]; then
```
myoung34, line 72: keep this

entrypoint.sh
myoung34: I think it makes sense to add 2 new vars total:

`CONFIGURED_ACTIONS_RUNNER_FILES_DIR` as you have, as well as `RUNNER_RETAIN_ENABLE` which is the behavior flag as true/false

myoung34 (changes requested):

myoung34:
@NickVeld i made a PR into this branch on your side with a suggestion on a refactor which mostly allows opt-in

entrypoint.sh
NickVeld: As I answered:

> "Opt-out" is a quite simple: do not mount a directory and a state will stay inside.

Could you please provide a use case when you mount the directory with the state but disables any logic of communicating with this mounted directory and the state.

If there is no such a use case then according to "Occam's razor" only one var is enough

entrypoint.sh
NickVeld: Yeah, makes sense

entrypoint.sh
NickVeld: Ok

entrypoint.sh
NickVeld: See the thread with the same topic above

entrypoint.sh
NickVeld: Moreover, I do not know what happens if try to configure when `/actions-runner/.runner` is presented. I do not think it is a good idea.

So this this particular line just checks that there is a somehow configured runner or not and not decide to retain the state or not.

myoung34: Implicit vs explicit for opt-in

The presense of a directory for a behavior change is implicit vs explicitly saying they want the behavior

If we want to stick with a single variable, then it should not have a default value, and the Null vs value can be used for opt in

NickVeld: opt-in:
```
ABC...
volumes:
      ## For avoiding reregistering
      - '${LOCAL_DIRECTORY_FOR_ACTIONS_RUNNER_FILES}:${CONFIGURED_ACTIONS_RUNNER_FILES_DIR}'
      - DEF...
```

opt-out:
```
ABC...
volumes:
      - DEF...
```
or
```
ABC...
volumes:
      ## For avoiding reregistering
      #- '${LOCAL_DIRECTORY_FOR_ACTIONS_RUNNER_FILES}:${CONFIGURED_ACTIONS_RUNNER_FILES_DIR}'
      - DEF...
```

invalid (using null):
```
ABC...
volumes:
      ## For avoiding reregistering
      - '${LOCAL_DIRECTORY_FOR_ACTIONS_RUNNER_FILES}:
      - DEF...
```

As for me, I think that mounting is quite explicit: You mount the state directory - you use the state management; you do not mount the state directory - you do not use state management.

In turn, a user when two variables:
> - I enabled `RUNNER_RETAIN_ENABLE` from supported envvar list
> - The system says `cp:  cannot stat '/actions-runner-files': No such file or directory`
> - ???

or

> - I enabled `RUNNER_RETAIN_ENABLE` from supported envvar list
> - The system says `You do not mount the directory with the state`
> - What is the directory ?

or

> - I mounted the directory with state
> - (The container worked some time) Strange... The state was not saved, why?
> - (Best case) Oh, it appears that I must/I forgot enable `RUNNER_RETAIN_ENABLE`.
> - (Worst case) This thing does not work. I do not have to research or asks for help so I will try another solution

entrypoint.sh
```diff
@@ -51,27 +51,51 @@ case ${RUNNER_SCOPE} in
...
-  trap deregister_runner SIGINT SIGQUIT SIGTERM
 fi
```
myoung34, line 96: we still want to keep the `_DISABLE_AUTOMATIC_DEREGISTRATION` logic here outside of the if else

myoung34 (changes requested):

entrypoint.sh
NickVeld: Yeah, I changed it locally and wait for you comments in order to decide whether to merge your PR or push only this change

NickVeld: I commented your PR. As we finish with the additional envvar discussion, I will merge your PR

myoung34 (approved):

myoung34:
I'm actually going on PTO and will be MIA for an extended period. we can merge this as is, theres only a few nuances I can sync back on so that we can get it in as-is

NickVeld:
Ok, I see, thanks

PR #130, opened 2021-08-03
Use platform-independent docker-compose instructions in base image

### LINKED ISSUE #128: Dockerfile.base fails to build on arm
Dockerfile.base has a provision to skip installation of docker-compose on arm.  However, it then unconditionally tries to chmod that file:

https://github.com/myoung34/docker-github-actions-runner/blob/29cbe17895816370a7d5ff1275fb81514d6a8cb8/Dockerfile.base#L58-L59

On my arm64 device (Nvidia Jetson Nano), the image therefore fails to build.

We could make the chmod part of the previous command with curl, or make it conditional on the file existing first, but I have another solution to offer.

I have found that upgrading pip and then using pip to install docker-compose works well on all platforms.

I will have a PR for this shortly.  Let me know what you think!

joeyparrish:
These work for me on all platforms, including arm64, lifting the
limitation on using docker-compose in arm images.

This also fixes a build error for me for the base image, caused when we
tried to unconditionally chmod docker-compose.

Closes #128

### DIFF
--- a/Dockerfile.base
+++ b/Dockerfile.base
@@ -37,6 +37,7 @@ RUN apt-get update && \
     openssh-client \
     locales \
     python3-pip \
+    python3-dev \
     dumb-init \
   && pip3 install --no-cache-dir awscliv2 \
   && locale-gen en_US.UTF-8 \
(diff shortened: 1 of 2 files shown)


myoung34:
To be honest I don't imagine a lot of people in the niche scenario of using docker-compose here (considering using docker from within is broken) need to use the python installed version (which does not get security updates among other things)

I'll leave this open, and if there are enough upvotes i'll look into it, but I'm not sure it's worth the changeset to be honest (installing from apt is best case scenario in most OS)

myoung34:
outdated for #133

PR #131, opened 2021-08-03
Make python3 the default version of python

### LINKED ISSUE #129: Failures with both /usr/bin/python and /usr/bin/env python
The base image appears to only install python3.  I don't disagree with that, but it doesn't get marked as the default version of python, so both `/usr/bin/python` and `/usr/bin/env python` fail in the shebang line of scripts used by my workflows.

I considered changing my workflow to explicitly install python2 or to explicitly set the default to python3, but I think it's best to make the runner's image more compatible with the default images used by GitHub.

joeyparrish:
The base image only installs python3, not python2.  But /usr/bin/env
python should still work, so now we make python3 the default version.

Closes #129

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # hadolint ignore=DL3007
-FROM myoung34/github-runner-base:latest
+FROM joeyparrish/github-runner-base:latest
 LABEL maintainer="myoung34@my.apsu.edu"
 
 ENV AGENT_TOOLSDIRECTORY=/opt/hostedtoolcache
--- a/Dockerfile.base
+++ b/Dockerfile.base
@@ -38,6 +38,7 @@ RUN apt-get update && \
     locales \
     python3-pip \
     dumb-init \
+  && update-alternatives --install /usr/bin/python python /usr/bin/python3 10 \
   && pip3 install --no-cache-dir awscliv2 \
   && locale-gen en_US.UTF-8 \
   && dpkg-reconfigure locales \


Dockerfile
```diff
@@ -1,5 +1,5 @@
 # hadolint ignore=DL3007
-FROM myoung34/github-runner-base:latest
+FROM joeyparrish/github-runner-base:latest
```
myoung34, line 2: this would need to be removed

Dockerfile.base
```diff
@@ -38,6 +38,7 @@ RUN apt-get update && \
     locales \
     python3-pip \
     dumb-init \
+  && update-alternatives --install /usr/bin/python python /usr/bin/python3 10 \
```
myoung34, line 41: in the python community its generally accepted that `python` is whatever the OS decides is default and `$ python3` is specific to python 3. Linking the two is a bad practice as python2 is still generally used (albeit not under best cases) and I don't intend to introduce that confusion here

myoung34 (changes requested):

myoung34:
The best practice is to be specific with python vs python3 executables

Dockerfile
joeyparrish: Oops!  Should have reviewed my own commits.

Dockerfile.base
joeyparrish: Understood.  Thanks for taking a look.

PR #133, opened 2021-08-04
Fix failing chmod on docker-compose on arm

### LINKED ISSUE #128: Dockerfile.base fails to build on arm

joeyparrish:
Since docker-compose is not downloaded on arm, the chmod command fails
there.  Moving it inside the previous command, after curl, ensures it
only runs when it should.

Closes #128


myoung34:
Thanks!

PR #135, opened 2021-08-04
Fix runner deregistration with ACCESS_TOKEN

joeyparrish:
This fixes runner deregistration, which needs to have ACCESS_TOKEN still
defined to fetch a fresh RUNNER_TOKEN for deregistration.

The original purpose of unsetting the variables would seem to be for
security, to make sure they unavailable to the jobs running in the
container.  Unexporting the variables achieves the same goal, but leaves
them available to the entrypoint later when it runs the deregistration
process.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -3,10 +3,17 @@
 export RUNNER_ALLOW_RUNASROOT=1
 export PATH=$PATH:/actions-runner
 
+# Un-export these, so that they must be passed explicitly to the environment of
+# any command that needs them.  This may help prevent leaks.
+export -n ACCESS_TOKEN
+export -n RUNNER_TOKEN
+
 deregister_runner() {
   echo "Caught SIGTERM. Deregistering runner"
-  _TOKEN=$(bash /token.sh)
-  RUNNER_TOKEN=$(echo "${_TOKEN}" | jq -r .token)
+  if [[ -n "${ACCESS_TOKEN}" ]]; then
+    _TOKEN=$(ACCESS_TOKEN="${ACCESS_TOKEN}" bash /token.sh)
+    RUNNER_TOKEN=$(echo "${_TOKEN}" | jq -r .token)
+  fi
   ./config.sh remove --token "${RUNNER_TOKEN}"
   exit
 }
@@ -54,7 +61,7 @@ esac
 configure_runner() {
   if [[ -n "${ACCESS_TOKEN}" ]]; then
     echo "Obtaining the token of the runnet"
-    _TOKEN=$(bash /token.sh)
+    _TOKEN=$(ACCESS_TOKEN="${ACCESS_TOKEN}" bash /token.sh)
     RUNNER_TOKEN=$(echo "${_TOKEN}" | jq -r .token)
   fi
 
@@ -100,9 +107,6 @@ fi
 if [[ ${_DISABLE_AUTOMATIC_DEREGISTRATION} == "false" ]]; then
   trap deregister_runner SIGINT SIGQUIT SIGTERM
 fi
-
-unset ACCESS_TOKEN
-unset RUNNER_TOKEN
 
 # Container's command (CMD) execution
 "$@"


myoung34:
Have you tested this? Can it deregister using a new token?

joeyparrish:
It worked for me, whereas it didn't (couldn't AFAICT) before.

First, you unset RUNNER_TOKEN before running CMD.  If that's the case, the trap can't reuse the same token once CMD is running.

Then in deregister_runner, you already ignore any previous value of RUNNER_TOKEN and run /token.sh to get a new one.  Except you can't, because /token.sh uses ACCESS_TOKEN, and you unset ACCESS_TOKEN, too.

My initial plan was to register with a prefix and random suffix, then deregister after each job.  Ultimately, I wound up settling on a different architecture for my service that doesn't involve deregistering, so I don't really need this myself.  But I thought this was still a useful improvement someone might benefit from.

entrypoint.sh
```diff
@@ -3,10 +3,17 @@
 export RUNNER_ALLOW_RUNASROOT=1
 export PATH=$PATH:/actions-runner
 
+# Un-export these, so that they must be passed explicitly to the environment of
+# any command that needs them.  This may help prevent leaks.
+export -n ACCESS_TOKEN
+export -n RUNNER_TOKEN
```
myoung34, line 9: this works in theory but not via docker, probably because the cmd is not a separate process (retains the same pid)?

I built this and ran it with 

```bash
docker run \
  -e REPO_URL="https://github.com/OctoKode/test1" \
  -e RUNNER_NAME="foo-runner" \
  -e RUNNER_TOKEN="AAGW...snip....WF6Q" \
  wut
```

then docker exec'd into the container:

```bash
root@f5e6653bcb76:/actions-runner# env | grep -i token
RUNNER_TOKEN=AAGW...snip....WF6Q
```

myoung34 (changes requested):

entrypoint.sh
myoung34: Actually looks like it doesnt work on my master branch either 🤔

myoung34 (changes requested):
.

entrypoint.sh
```diff
@@ -3,10 +3,17 @@
...
+
 deregister_runner() {
   echo "Caught SIGTERM. Deregistering runner"
-  _TOKEN=$(bash /token.sh)
-  RUNNER_TOKEN=$(echo "${_TOKEN}" | jq -r .token)
+  if [[ -n "${ACCESS_TOKEN}" ]]; then
+    _TOKEN=$(ACCESS_TOKEN="${ACCESS_TOKEN}" bash /token.sh)
+    RUNNER_TOKEN=$(echo "${_TOKEN}" | jq -r .token)
+  fi
   ./config.sh remove --token "${RUNNER_TOKEN}"
```
myoung34, line 17: this will never actually run because we're unsetting RUNNER_TOKEN above even if the temp token is still valid

myoung34: perhaps put this line in the if condition and add:

```
else
  echo ACCESS_TOKEN variable is not set and required to deregister
```

myoung34 (changes requested):

myoung34:
@joeyparrish just following up

joeyparrish:
If you're not interested in this, or don't see the benefit, let's just close it.  It's not important to me any more.

myoung34:
I've been on vacation. Feel free to reopen, the idea itself is fine I just havent had time to circle back.

joeyparrish:
You're welcome to the code, but I am not using it myself any more, so I'm going to move on from this PR.  Thanks!

myoung34:
@joeyparrish Im not able to reopen this for some reason. It might resolve #139

joeyparrish:
I'll recreate it from a branch this time (instead of my fork's master branch).  Unfortunately, I've already overwritten the master branch of my fork.

PR #145, opened 2021-09-13
Fix #142 - building on Sid

### LINKED ISSUE #142: Debian Sid failing
GitHub Actions Sid builds are failing because Sid is apparently not on their repository (anymore?) https://download.docker.com/linux/debian/dists/
I said that I had Sid running in https://github.com/myoung34/docker-github-actions-runner/pull/123#pullrequestreview-711525427 so I cannot imagine it not working before. Maybe they stopped providing Sid packages because no one in their right mind would use Docker on Sid?

Here is a GHA failing on Sid https://github.com/myoung34/docker-github-actions-runner/runs/3569202962?check_suite_focus=true

Niek:
Added `distro` and `version` variables to use in the apt repo config, so it's better future-proof as well.

### DIFF
--- a/Dockerfile.base
+++ b/Dockerfile.base
@@ -51,9 +51,14 @@ RUN apt-get update && \
   && make \
   && make install \
   && cd / \
+  # Determine the Distro name (Debian, Ubuntu, etc)
+  && distro=$(lsb_release -is | awk '{print tolower($0)}') \
+  # Determine the Distro version (bullseye, xenial, etc)
+  # Note: sid is aliased to bullseye, because Docker doesn't have a matching apt repo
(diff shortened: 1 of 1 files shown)


myoung34 (approved):

myoung34:
verified it builds successfully with: 

```
$ cp Dockerfile.base Dockerfile.base.debian-sid; sed -i.bak "s/FROM.*/FROM debian:sid/" Dockerfile.base.debian-sid
$ docker build -f Dockerfile.base.debian-sid -t foo .
$ cp Dockerfile Dockerfile.debian-sid; sed -i.bak "s/FROM.*/FROM foo/" Dockerfile.debian-sid
$ docker build -f Dockerfile.debian-sid -t wut .
```

Thanks @Niek

PR #150, opened 2021-09-22
Fix small typo in entrypoint.sh log output

gazab:
Extremely small and practically pointless typo fix. But I just saw it in the output and it annoyed me a bit 😃

image


myoung34:
thanks!

PR #161, opened 2021-11-24
Support --ephemeral option

Kesin11:
fix: #148 

Add `EPHEMERAL` environment variable to support `--ephemeral` option.

@joeyparrish refers also need to fix ephemeral_runner.sh, but it seems no problem with using default ENTRYPOINT and CMD.
I checked using this docker run command.

```bash
docker run --rm --name github-runner \
  -e RUNNER_NAME_PREFIX="ephemeral-docker" \
  -e ACCESS_TOKEN="${RUNNER_PAT}" \
  -e RUNNER_SCOPE="org" \
  -e ORG_NAME="${ORG}" \
  -e LABELS="container" \
  -e EPHEMERAL=1 \
  -v /var/run/docker.sock:/var/run/docker.sock \
  kesin11/docker-github-actions-runner:latest
```

PR #162, opened 2021-11-27
Fix configure command when EPHEMERAL is disabled

Kesin11:
When run with no `EPHEMERAL` env, `Unrecognized command-line input arguments: 'replace'. For usage refer to: .\config.cmd --help or ./config.sh --help`  message appeared at entrypoint.sh .

When `EPHEMERAL` env is disabled, `--replace "${_EPHEMERAL}"` is replaced to `--replace ''` so it is treated as part of `--replace` option. This bug comes from #161.


myoung34 (approved):

Kesin11:
Thanks!

PR #177, opened 2022-01-24
Update AWS CLI to v2

lauripiispanen:
AWS no longer pushes the AWS CLI to PyPi and the version bundled with the runner is getting quite old. This PR pulls the latest AWS CLI v2 directly from AWS. As of this time, the official runners are running AWS CLI 2.4.12.


myoung34:
nice catch. Thanks!

PR #178, opened 2022-01-26
fix(ephemeral-runner): Replace once with ephemeral

MrKevinWeiss:
When running the `ephemeral-runner.sh` there is a deprecation warning.
Better to solve it sooner than later...
```
Warning: '--once' is going to be deprecated in the future, please consider using '--ephemeral' during runner registration.
```


myoung34 (approved):

myoung34:
Thanks!

PR #179, opened 2022-01-26
feat: add github cli (gh)

jarrett-confrey:
add github cli (gh) to closer align to github-hosted runners

https://github.com/cli/cli#github-actions
>GitHub CLI comes pre-installed in all GitHub-Hosted Runners.

installed using instructions from [install_linux.md](https://github.com/cli/cli/blob/trunk/docs/install_linux.md#debian-ubuntu-linux-raspberry-pi-os-apt) with sudo removed

### DIFF
--- a/Dockerfile.base
+++ b/Dockerfile.base
@@ -70,6 +70,11 @@ RUN echo en_US.UTF-8 UTF-8 >> /etc/locale.gen \
   && ( [[ $(lsb_release --codename | cut -f2) == "jammy" ]] && echo "Ubuntu Jammy is marked as beta. Please see https://github.com/actions/virtual-environments/issues/5490" || : ) \
   && rm -rf /var/lib/apt/lists/* \
   && rm -rf /tmp/* \
+  && curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg | dd of=/usr/share/keyrings/githubcli-archive-keyring.gpg \
+  && chmod go+r /usr/share/keyrings/githubcli-archive-keyring.gpg \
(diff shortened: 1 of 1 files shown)


myoung34:
Can you confirm this works for armv7?

I dont see it listed [here](https://github.com/cli/cli/releases)

Also, it might be easier to use something more closely resembling how the upstream installs it into their internal runners [similar to this](https://github.com/actions/virtual-environments/blob/main/images/linux/scripts/installers/github-cli.sh)

joeyparrish:
They list arm6 and arm64.  Would the arm6 builds run on arm7 devices?

Personally, I think using the official installation instructions makes a lot of sense.  The virtual-environment install script uses some magic internal helper functions.  Would you want those imported?

myoung34:
@jarrett-confrey can you confirm this still works with all the latest changes to Dockerfile.base ?
I dropped support for armv7 so I dont mind getting this in as a result if you don't mind confirming that it will work as expected for you

Dylan-Feith:
Hi @myoung34 
we are interesting in this feature too.
What was commited by @jarrett-confrey is exactly what can be found in gh documentation  : 
https://github.com/cli/cli/blob/trunk/docs/install_linux.md

It is written in the doc that commands are compatible with :  i386, amd64, arm64, armhf.

Please can we solve this feature request ?

jarrett-confrey:
@myoung34 i have rebased and tested, it seems to still work as expected

myoung34:
Thanks!

Dylan-Feith:
Thank you, I have deployed latest container version and it worked perfectly :ok_hand:

PR #180, opened 2022-01-27
fix(README.md): Fix the service example

MrKevinWeiss:
I was able to get everything up and running and it looks good aside from a small gotcha.

While setting up an ephemeral systemd service it seemed that I still needed to mount the `/var/run/docker.sock`.

I am a pretty big fan of copy pasta and this may save a few google queries for "cannot connect to the docker daemon at unix:///var/run/docker.sock. is the docker daemon running?"

I guess some issue with docker in docker.

Anyways, just wondering if it would be useful, no need to merge if it doesn't make sense.


myoung34 (approved):

myoung34:
Thanks!

PR #188, opened 2022-02-17
Add proxy env variable passthru on sudo

ikari7789:
If you use "sudo" in the workflow file while behind a proxy, the environment variables passed into the container on creation will be lost since sudo current resets the environment. Update /etc/sudoers to allow the *_proxy and *_PROXY variables to be passed thru so proxy configuration is not lost.

### DIFF
--- a/Dockerfile.base
+++ b/Dockerfile.base
@@ -13,6 +13,7 @@ SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 ENV DEBIAN_FRONTEND=noninteractive
 # hadolint ignore=DL3003,DL4001,SC2086
 RUN echo en_US.UTF-8 UTF-8 >> /etc/locale.gen \
+  && sed -e 's/Defaults	env_reset/Defaults	env_keep = "HTTP_PROXY HTTPS_PROXY NO_PROXY FTP_PROXY http_proxy https_proxy no_proxy ftp_proxy"/' -i /etc/sudoers
   && apt-get update \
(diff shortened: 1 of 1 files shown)


myoung34 (approved):

myoung34:
Thanks!

myoung34:
@ikari7789 I had to revert this. 

* there's a typo on the line, it needs to end with a `\`
* this doesnt work as-is: 

```
$ docker run -it ubuntu:focal /bin/bash
root@4a914d509cab:/# ls /etc/sudoers
ls: cannot access '/etc/sudoers': No such file or directory
```

Dockerfile.base
```diff
@@ -13,6 +13,7 @@ SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 ENV DEBIAN_FRONTEND=noninteractive
 # hadolint ignore=DL3003,DL4001,SC2086
 RUN echo en_US.UTF-8 UTF-8 >> /etc/locale.gen \
+  && sed -e 's/Defaults	env_reset/Defaults	env_keep = "HTTP_PROXY HTTPS_PROXY NO_PROXY FTP_PROXY http_proxy https_proxy no_proxy ftp_proxy"/' -i /etc/sudoers
```
myoung34, line 16: This needs to end with a `\`

ikari7789: Totally my fault. This is what I get for copy and pasting from my local extended image. Sorry!

PR #192, opened 2022-03-13
add JOBS_ACCEPTANCE_TIMEOUT

sakamoto66:
I'm using a runner with the `--ephemeral` option.
I thought of this option to automatically terminate Runners who did not receive a job for some reason.

> Optional environment variable to terminate the runner when the job is not received for a certain period of time. Set the number of seconds to time out. If the job is not accepted within the specified number of seconds, the runner will exit. Defaults to `""`.


myoung34:
I'll have to think on this one. My gut says that it kind of goes against what most schedulers (nomad, k8s, etc) will want and will provide some behavior nuances

myoung34:
@sakamoto66 Im not against this PR per-se, but I do have a high level question:

How does this fit with docker?
Im curious how you plan to choreograph a runner that will continuously reboot itself.

In all schedulers it will always want to run `n`. If this time out hits it will simply stop and restart.

The only scenario I can imagine is using custom metrics to scale a runner, but that logic should not live within the container itself. I cant think of any software that kills itself on timeouts, that would just be done at the orchestrator level with a metric of number of containers based on pending jobs, etc.

sakamoto66:
@myoung34 
Sorry for my late reply.

It is assumed that Runner will be started and used only when necessary.
I wanted this feature when I did AWS Batch.

Imagine a situation where you specify --ephemeral and set the batch timeout to 1 hour.
If Runner is used, Docker will close as expected after the process is completed, so there is no problem.
If Runner is not used, it will wait for an hour without doing anything.
To avoid this waste, I thought it would be nice to have an acceptance timeout separate from the batch timeout.

myoung34:
Wouldnt ephemeral make sure that the node that scaled up got the job though? Thats my understanding from the docs

entrypoint.sh
```diff
@@ -118,5 +118,9 @@ fi
 extra_flags=""
 [[ -n "$DISABLE_AUTO_UPDATE" ]] && extra_flags="--disableupdate" || :
 
+if [ -n "${JOBS_ACCEPTANCE_TIMEOUT}" ]; then
```
```suggestion
if [[ -n "${JOBS_ACCEPTANCE_TIMEOUT}" ]]; then
```

check_jobs.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+# Check jobs acceptance timeout.
+# If jobs cannot be confirmed, listening will be terminated.
+
+if [ -z "$1" ]; then exit 1; fi
```
myoung34: this can never happen since we already ran `-n` on that env var which is passed here

```suggestion

```

check_jobs.sh
```diff
@@ -0,0 +1,20 @@
...
+
+_RUNNER_WORKDIR=${RUNNER_WORKDIR:-/_work}
+_JOBS_ACCEPTANCE_TIMEOUT=$1
+
+while [ 1 ]; do
+  if [ ! -e ${_RUNNER_WORKDIR}/_temp/_github_workflow ]; then
```
myoung34: i dont follow what this is doing

entrypoint.sh
```diff
@@ -118,5 +118,9 @@ fi
...
+  /check_jobs.sh ${JOBS_ACCEPTANCE_TIMEOUT} &
+fi
+
 # Container's command (CMD) execution
 "$@" "${extra_flags}"
```
myoung34: after you pkill the listener this will run regardless meaning it will spin up after it was asked to quit

myoung34 (changes requested):

entrypoint.sh
sakamoto66: I tried to check if Runner.Listener is running on check_jobs.sh.

check_jobs.sh
sakamoto66: Changed "1" to process check.
Changed check directory, _github_workflow to log directory.

sakamoto66:
Hi, @myoung34 
Sorry for delay. It took me a while to think of an idea.
I want you to see the code.

myoung34:
After sitting on this for way too long (Absolutely my fault) I dont think Ill merge this for a few reasons:

* SRP - this container isnt about scheduling in any way - this is a scheduler concern (See [arc](https://github.com/actions/actions-runner-controller) for example)
* Complexity - scheduling (above) introduces some complexities about what it means to "not accept a job" and this feels more like an operator. this puts a weird amount of non-intuitive logic into something whose only job is to run actions.

Again I apologize for sitting on this so long.

PR #195, opened 2022-04-02
Fix disable auto update

### LINKED ISSUE #191: DISABLE_AUTO_UPDATE flag no longer respected
I found our containers a in a "self upgrade restart loop" again this morning, although we're using the DISABLE_AUTO_UPDATE flag. It worked just fine for the past 3 weeks until now.

Can someone confirm?

myoung34:
Resolves #191 

@Kesin11 is correct. The flag was given to the wrong part of the process.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -72,6 +72,13 @@ configure_runner() {
     _EPHEMERAL=""
   fi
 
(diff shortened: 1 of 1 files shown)


Kesin11:
Thank you for fix it!

PR #198, opened 2022-04-14
Add Nomad autoscaling example

smaeda-ks:
Hi there, I've been impressed by this awesome image and really thankful for that.

After playing around with this image, I recently came up with the idea of autoscaling runners using Nomad and ephemeral runners relatively easily, so I'd like to add some notes to give folks this basic idea to scale their deployment model even further.

Let me know if you have any feedback/concerns, thank you!

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,6 +204,10 @@ job "github_runner" {
 }
 ```
 
+With Nomad, it's also easy to achieve an "on-demand" Self-hosted Runner deployment model that auto-scales horizontally, which most high-load CI/CD pipelines should take into account to save operational costs. See [smaeda-ks/nomad-github-runners-autoscaler-demo](https://github.com/smaeda-ks/nomad-github-runners-autoscaler-demo) for more details.
+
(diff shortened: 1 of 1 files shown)


myoung34:
As much as I like this I'm more concerned with examples of how to "just run" them here

Architectural patterns would add a lot more noise with all the variations (k8s, nomad, ecs, etc)

I think you'd be in a good spot to start a new repo though that uses this runner to show those patterns in different orchestrators though and probably get linked to from the awesome-actions repo if that's appealing

smaeda-ks:
@myoung34 Thanks for your comment! While I understand your point, I don't think this would add much noise as you think. Conversely, this will add more value to users who don't know how to approach that path, or folks trying to adopt it for production use, increasing their attention to this image. Autoscaling runners is such a critical component of the production system, and if I don't see any possibility to achieve that using this image in the first place I might just leave the page. This tiny addition would advocate this image more than its textual noise, I believe.

smaeda-ks:
Or, if your concern is more like avoiding README getting too long text, we could have a dedicated "examples" folder and move each platform-specific example to there, and replace them with those links.

myoung34:
I think maybe a wiki might be more appropriate

Let me think on this for a few days/over the weekend

smaeda-ks:
Much appreciated :) Let me know if there's anything I can help you with on this one, I'm more than happy to assist.

PR #205, opened 2022-04-29
Support authentication via Github App

AliAllomani:
Adding support for authentication via Github Apps


myoung34:
I'd like to avoid mixing and matching this much bash and python. If the JWT can be done with bash i say go for it

myoung34 (changes requested):
I'd like to avoid mixing and matching this much bash and python. If the JWT can be done with bash i say go for it

AliAllomani:
it can be done but it's going to be much more complicated with bash.

In this PR I'm using python for API calls to github, and the remaining runner's config is kept as bash, so in my opinion i see it as fair amount of mixing.

Another suggestion would be reducing bash code by calculating all the required params for `./config.sh` in the python script and pass them to it.

samsen1:
This can definitely be done w/ bash. we do it now.

```
#!/usr/bin/env bash

# Generate JWT for Github App
#
# Credit: 
#   https://gist.github.com/carestad/bed9cb8140d28fe05e67e15f667d98ad 

thisdir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
set -o pipefail

# Change these variables:
app_id=<APP_ID>
app_private_key="${PEM_KEY}"

# Shared content to use as template
header='{
    "alg": "RS256",
    "typ": "JWT"
}'
payload_template='{}'

build_payload() {
        jq -c \
                --arg iat_str "$(date +%s)" \
                --arg app_id "${app_id}" \
        '
        ($iat_str | tonumber) as $iat
        | .iat = $iat
        | .exp = ($iat + 300)
        | .iss = ($app_id | tonumber)
        ' <<< "${payload_template}" | tr -d '\n'
}

b64enc() { openssl enc -base64 -A | tr '+/' '-_' | tr -d '='; }
json() { jq -c . | LC_CTYPE=C tr -d '\n'; }
rs256_sign() { openssl dgst -binary -sha256 -sign <(printf '%s\n' "$1"); }

sign() {
    local algo payload sig
    algo=${1:-RS256}; algo=${algo^^}
    payload=$(build_payload) || return
    signed_content="$(json <<<"$header" | b64enc).$(json <<<"$payload" | b64enc)"
    sig=$(printf %s "$signed_content" | rs256_sign "$app_private_key" | b64enc)
    generated_jwt="${signed_content}.${sig}"

    app_installations_url="https://api.github.com/app/installations"
    app_installations_response=$(curl -sX GET -H "Authorization: Bearer  ${generated_jwt}" -H "Accept: application/vnd.github.v3+json" ${app_installations_url})
    access_token_url=$(echo $app_installations_response | jq '.[] | select (.app_id  == '${app_id}') .access_tokens_url' --raw-output)

    access_token_response=$(curl -sX POST -H "Authorization: Bearer  ${generated_jwt}" -H "Accept: application/vnd.github.v3+json" ${access_token_url})

    export ACCESS_TOKEN=$(echo $access_token_response | jq .token --raw-output)
}

sign
```

seaniag:
For me this is a critical feature as my security requirements don't like setting access tokens as an env var. The image for this is already huge so I see no problem with having some python as well. Its far more readable than the bash anyway.

Uzlopak:
@samsen1 
If you provide a separate PR i assume it can be reviewed and potentially merged.

lina-clari:
Can we please add this functionality as soon as possible? could be very beneficial. This seems to be a bash implementation https://github.com/michaelsync/github-actions-self-hosted-runner-container/blob/ae6c023f0985d1b5d35731f592d2abe577a86b98/entrypoint.sh#L39-L61 ( have not tested it ... ).

davidkarlsen:
If you use Kubernetes, this is possible by using this operator: https://github.com/evryfs/github-actions-runner-operator

myoung34:
Closing in favor of #254

PR #206, opened 2022-05-09
option --ephemeral not supported with /actions-runner/run.sh

pitpit:
when running `myoung34/github-runner:latest` with command set to `/ephemeral-runner.sh`, the following error occurs:

```
Unrecognized command-line input arguments for command run: 'ephemeral'. For usage refer to: .\config.cmd --help or ./config.sh --help
```

`--ephemeral` is passed to config.sh in the entrypoint

### DIFF
--- a/ephemeral-runner.sh
+++ b/ephemeral-runner.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 echo "*** Starting ephemeral runner. ***"
-/actions-runner/run.sh --ephemeral
+/actions-runner/run.sh
 rv=$?
 
 # See exit code constants in the runner source here:


myoung34:
@pitpit in hindsight: do we need ephemeral-runner.sh anymore?

I'm drawing a blank on context but iirc @joeyparrish added this to better handle what happens with update loops
In fact I'm starting to face confusion on non-standard behavior like what ephemeral-runner.sh does and what is requested in #192 

Would it make sense to combine all this into a better `ephemeral-runner.sh` now that auto-update wont happen?

joeyparrish:
I think it's no longer needed.  I keep forgetting to revisit this repo and remove it.

Here's my current setup for self-hosted runners in my team's test lab, with some details like ports and bind mounts omitted. 
 We're using something like this systemd unit:

```
[Unit]
Description=Ephemeral GitHub Actions Runner Container
After=docker.service
Requires=docker.service

[Service]
TimeoutStartSec=0
Restart=always
ExecStartPre=-/usr/bin/docker stop %N
ExecStartPre=-/usr/bin/docker rm %N
ExecStartPre=-/usr/bin/docker pull myoung34/github-runner:latest
ExecStart=/usr/bin/docker run --rm --env-file /etc/ephemeral-github-actions-runner.env -e RUNNER_NAME=%H --name %N myoung34/github-runner:latest

[Install]
WantedBy=multi-user.target
```

Our env file (with some details omitted) looks like:

```sh
RUNNER_SCOPE=org
ORG_NAME=*****

#RUNNER_SCOPE=repo
#REPO_URL=https://github.com/*****/*****

LABELS=*****
ACCESS_TOKEN=*****
RUNNER_WORKDIR=/tmp/runner/work
DISABLE_AUTO_UPDATE=1
EPHEMERAL=1
```

This is working very well for us now without any alternate entrypoint.  I think the script can be dropped and the docs updated.  I'll send a PR shortly.

PR #211, opened 2022-06-02
Upgrade docker image to latest LTS version (jammy)

Ru13en:
Podman, buildah and skopeo are only available at official ubuntu repro in jammy distro.
Otherwise you need to add deprecated kubic repro


myoung34:
Unfortunately a lot of things need to change to update the base LTS including rolling down all supported ones. The readme, GitHub actions, etc

Ru13en:
Right, but it only affects actions requiring software that you don't have in the base image.
Check out it:
https://github.com/actions/virtual-environments/issues/5490

Ru13en:
@myoung34 please, check if everything is ok.
It should be possible to release all versions (bionic, focal, and jammy) with GitHub actions...

.github/workflows/base.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       - name: Copy Repo Files
         uses: actions/checkout@master
       - name: Get GitHub organization or user
-        run: echo 'ORG='$(dirname ${GITHUB_REPOSITORY}) >> $GITHUB_ENV
+        run: echo 'ORG='$(echo $(dirname ${GITHUB_REPOSITORY}) | awk '{print tolower($0)}') >> $GITHUB_ENV
```
myoung34, line 20: whats the use-case for lower-casing the org here? docker tags dont care as far as i know

Dockerfile.base
```diff
@@ -13,11 +11,10 @@ SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 ENV DEBIAN_FRONTEND=noninteractive
 # hadolint ignore=DL3003,DL4001,SC2086
 RUN echo en_US.UTF-8 UTF-8 >> /etc/locale.gen \
-  && apt-get update \
-  && apt-get install -y --no-install-recommends gnupg \
-  && apt-key adv --keyserver keyserver.ubuntu.com --recv-keys ${GIT_CORE_PPA_KEY} \
```
myoung34, line 18: This all needs to stay. There are wild differences in git versions between OS versions:

```
Ubuntu:bionic default 2.17.1 -> 2.36.1
ubuntu:Focal default 2.25.1 -> 2.36.1
ubuntu:Jammy default 2.34.1 -> 2.36.1
debian:Bullseye default 2.30.2 -> 2.36.1
debian:Sid default 2.36.1 -> 2.36.1
```

Dockerfile.base
```diff
@@ -40,27 +37,25 @@ RUN echo en_US.UTF-8 UTF-8 >> /etc/locale.gen \
...
   && ( [[ $(apt-cache search -n liblttng-ust1 | awk '{print $1}') == "liblttng-ust1" ]] && apt-get install -y --no-install-recommends liblttng-ust1 || : ) \
   && ( ( curl "https://awscli.amazonaws.com/awscli-exe-linux-$(uname -m).zip" -o "awscliv2.zip" && unzip awscliv2.zip -d /tmp/ && /tmp/aws/install ) || pip3 install awscli ) \
   # Determine the Distro name (Debian, Ubuntu, etc)
   && distro=$(lsb_release -is | awk '{print tolower($0)}') \
   # Determine the Distro version (bullseye, xenial, etc)
   # Note: sid is aliased to bullseye, because Docker doesn't have a matching apt repo
-  && apt-key adv --keyserver keyserver.ubuntu.com --recv-keys ${DOCKER_KEY} \
-  && curl -fsSL https://download.docker.com/linux/${distro}/gpg | apt-key add - \
+  && mkdir -p /etc/apt/keyrings \
+  && ( curl -fsSL https://download.docker.com/linux/${distro}/gpg | gpg --dearmor -o /etc/apt/keyrings/docker.gpg ) \
```
myoung34, line 58: whats the benefit of this over `apt-key adv` ?

Dockerfile.base
```diff
@@ -40,27 +37,25 @@ RUN echo en_US.UTF-8 UTF-8 >> /etc/locale.gen \
...
+    git \
     dumb-init \
-    podman \
-    buildah \
-    skopeo \
+    nodejs \
   && sed -e 's/Defaults.*env_reset/Defaults env_keep = "HTTP_PROXY HTTPS_PROXY NO_PROXY FTP_PROXY http_proxy https_proxy no_proxy ftp_proxy"/' -i /etc/sudoers \
-  && echo deb http://ppa.launchpad.net/git-core/ppa/ubuntu $([[ $(grep -E '^ID=' /etc/os-release | sed 's/.*=//g') == "ubuntu" ]] && (grep VERSION_CODENAME /etc/os-release | sed 's/.*=//g') || echo bionic) main>/etc/apt/sources.list.d/git-core.list \
-  && apt-get update \
-  && ( apt-get install -y --no-install-recommends git || apt-get install -t stable -y --no-install-recommends git || apt-get install -y --no-install-recommends git=1:2.33.1-0ppa1~ubuntu18.04.1 git-man=1:2.33.1-0ppa1~ubuntu18.04.1 ) \
```
myoung34: read above, this is crucial for git versioning

myoung34 (changes requested):
Also please squash it all down if you get a chance

.github/workflows/base.yml
Ru13en: It cares a lot. My github username has the first letter uppercase and if I didn't add it, the CI failed in my repro.

Dockerfile.base
Ru13en: If so, you cannot use the PPA_KEY set in the dockerfile for Jammy, since it's deprecated. 
You must change it to the new gpg key.

Dockerfile.base
Ru13en: You don't need to set the Key id... 
it makes it for you when you make "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.gpg]..."

Dockerfile.base
myoung34: Deprecations are still fine for the time being, it will still work and exit 0

Dockerfile.base
Ru13en: Re-added

Dockerfile.base
Ru13en: ok, for git I have re-added it, but for docker, I'm using the gpg key file as the recommended way by the docker installation procedure

.github/workflows/base.yml
myoung34: I could have sworn that wouldnt happen but I just saw your failure: `Error: buildx failed with: error: failed to solve: failed to parse stage name "Ru13en/github-runner-base:ubuntu-bionic": invalid reference format: repository name must be lowercase`

Wow...thats really lame.

👍

Ru13en:
Squash made, git back to ppa repros. I think everything is ok since my actions didn't fail to build all versions...

Dockerfile.base
```diff
@@ -55,12 +54,14 @@ RUN echo en_US.UTF-8 UTF-8 >> /etc/locale.gen \
...
+  && ( [[ $(lscpu -J | jq -r '.lscpu[] | select(.field == "Vendor ID:") | .data') == "ARM" ]] && echo "Not installing docker-compose. See https://github.com/docker/compose/issues/6831" || ( curl -sL "https://github.com/docker/compose/releases/download/${DOCKER_COMPOSE_VERSION}/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose && chmod +x /usr/local/bin/docker-compose ) ) \
+  && ( [[ $(lsb_release --codename | cut -f2) == "focal" ]] && ( echo "deb https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_20.04/ /" | tee /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list && curl -L "https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_20.04/Release.key" | apt-key add - && apt-get update) || : ) \
+  && ( [[ $(lsb_release --codename | cut -f2) == "focal" || $(lsb_release --codename | cut -f2) == "jammy" ]] && apt-get install -y --no-install-recommends podman buildah skopeo || : ) \
```
myoung34: would you be willing to link to [this](https://github.com/actions/virtual-environments/issues/5490) as an echo? I will very much forget over time why this check exists

.github/workflows/base.yml
Ru13en: Yep, that's why I added this fix and also another for release builds to use the same base build from org repro.

Dockerfile.base
```diff
@@ -55,12 +54,14 @@ RUN echo en_US.UTF-8 UTF-8 >> /etc/locale.gen \
...
-  && [[ $(lscpu -J | jq -r '.lscpu[] | select(.field == "Vendor ID:") | .data') == "ARM" ]] && echo "Not installing docker-compose. See https://github.com/docker/compose/issues/6831" || ( curl -sL "https://github.com/docker/compose/releases/download/${DOCKER_COMPOSE_VERSION}/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose && chmod +x /usr/local/bin/docker-compose ) \
+  && ( [[ $(lscpu -J | jq -r '.lscpu[] | select(.field == "Vendor ID:") | .data') == "ARM" ]] && echo "Not installing docker-compose. See https://github.com/docker/compose/issues/6831" || ( curl -sL "https://github.com/docker/compose/releases/download/${DOCKER_COMPOSE_VERSION}/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose && chmod +x /usr/local/bin/docker-compose ) ) \
+  && ( [[ $(lsb_release --codename | cut -f2) == "focal" ]] && ( echo "deb https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_20.04/ /" | tee /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list && curl -L "https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_20.04/Release.key" | apt-key add - && apt-get update) || : ) \
```
myoung34, line 64: will this work for debian and non-jammy ubuntu forks?

Not sure we should make it _only_ focal unless we have to

myoung34 (changes requested):
A question and adding a link will get this merged. After this goes through Id like to refactor my docs a bit to clean up the README but I'll tackle that

Dockerfile.base
Ru13en: Kubic repro is discontinued as described [here](https://podman.io/getting-started/installation) and Focal is the first LTS where podman was introduced... So It seems that is not right to try to add podman to bionic builds.
Also, debian sid and Bullseye doesn't require kubic repro... So that's why I added this only for focal.

Dockerfile.base
Ru13en: Added :)

myoung34 (approved):

myoung34:
Thanks! Ill work on updating some docs and making sure the builds work as expected

Dockerfile.base
nmalaguti: We just had an issue with the kubic repo being down and blocking us from running `apt-get update`. Is there a reason we need `podman` installed inside the containers?

Ru13en: I use this Github runner container to build oci images with buildah actions and other redhat actions. 
Without podman inside the container, you cannot execute those actions...
Also, if you are blocked from kubic repo, it should be related to the host machine and not the container.

myoung34: Honestly the best path here is to remove any non-standard repos at the end of the build which would lower it to just canonical mirros being down. which would be a bigger issue than a few third parties

PR #212, opened 2022-06-06
Add rsync bin

tomgrv:
Add rsync to base image


myoung34:
@tomgrv I tried to resolve this conflict but cant. instead of making you did it I just went ahead and fixed it with #214 

This is easier to merge now as i've dropped support for armv7/armhf

cwilby:
Thank you both!

Issue #186, opened 2022-02-04
Change UID/GUID in the runner image


Otterian:
Hi,

Thank you for your time and continuing commitment to this project.

Ideally the user in the image should run as UID 1001 and GID 121, as it does in the official runners from GitHub. This in order to be as identical as possible to the official runners, and prevent possible issues with Actions that may be depending on these variables.

We've done this in our downstream derivative of the Ubuntu image: https://github.com/evryfs/github-actions-runner/blob/test-NewImage/Dockerfile#L23-L24 - and it'd be great if it is done here as well. In addition to the mentioned reasons for doing this - dropping root in images is good security practice in general - even _if_ downstream images drop root.

I'd be happy to create a PR for this if you want this implemented. :+1:

Please let me know if anything is unclear, or if you have other thoughts about this :smile:

myoung34:
If you can scaffold something I can see the value in adding it for sure

Skyxim:
For example, when compiling OpenWRT, using root will be warned

myoung34:
resolved by #223 

See [the usage docs here](https://github.com/myoung34/docker-github-actions-runner/wiki/Usage#non-root-runners)

PR #223, opened 2022-07-25
Run as non-root user by default

marcus-bcl:


Uzlopak:
Should we use also group 121? see #186 

https://github.com/evryfs/github-actions-runner/blob/25912cce1c94b7f4712b94769ccf6e2b9daab1e8/Dockerfile#L62

marcus-bcl:
> Should we use also group 121? see #186
> 
> [evryfs/github-actions-runner@`25912cc`/Dockerfile#L62](https://github.com/evryfs/github-actions-runner/blob/25912cce1c94b7f4712b94769ccf6e2b9daab1e8/Dockerfile#L62)

Good call, and thanks for linking your repo - very helpful!

Uzlopak:
I am just curious if there is a significance in the order of USER and ENTRYPOINT. Should we call first USER and then ENTRYPOINT to avoid that we maybe not run the entrypoint script with the prefererred user?

myoung34:
> I am just curious if there is a significance in the order of USER and ENTRYPOINT. Should we call first USER and then ENTRYPOINT to avoid that we maybe not run the entrypoint script with the prefererred user?

Yes and no. `USER` bakes in the user and also runs subsequent steps as that user.
If the only thing after `USER` is `CMD` and/or `ENTRYPOINT` then order doesnt matter since all 3 of those become changeable at run time and only run at run time. If there are `RUN` after `USER` then it does matter as those will be run at build time after the user has changed.

Dockerfile
```diff
@@ -19,6 +19,12 @@ RUN chmod +x /actions-runner/install_actions.sh \
 
 COPY token.sh entrypoint.sh /
 RUN chmod +x /token.sh /entrypoint.sh
-
 ENTRYPOINT ["/entrypoint.sh"]
```
myoung34: even though i answered outside the thread and order doesn't matter it's still best practice to keep ENTRYPOINT at the end especially after RUN and USER

myoung34:
```suggestion
```

Dockerfile
```diff
@@ -19,6 +19,12 @@ RUN chmod +x /actions-runner/install_actions.sh \
...
+
+RUN groupadd -g 121 runner \
+    && useradd -mr -d /home/runner -u 1001 -g 121 runner \
+    && usermod -aG docker runner \
+    && chown runner /_work/ /opt/hostedtoolcache/
+USER runner
+
 CMD ["./bin/Runner.Listener", "run", "--startuptype", "service"]
```
myoung34:
```suggestion
CMD ["./bin/Runner.Listener", "run", "--startuptype", "service"]
ENTRYPOINT ["/entrypoint.sh"]
```

myoung34 (changes requested):

Dockerfile
marcus-bcl: Sure, I've moved the `ENTRYPOINT` after the user creation.  I've also combined the two `RUN`s above to prevent Sider from complaining, but let me know if you'd rather keep them separate.

Uzlopak (approved):
LGTM

myoung34:
Code looks good and works as expected, the only issue I see (and not entirely sure how to resolve) is `RUNNER_WORKDIR` env var

Unfortunately the logs are useless (ie non-existent) in that scenario (see [here](https://github.com/OctoKode/test1/runs/7522093359?check_suite_focus=true) )

I ran this with `-e RUNNER_WORKDIR="/foo"` and got:

```
2022-07-26 14:13:57Z: Listening for Jobs
2022-07-26 14:14:27Z: Running job: test
2022-07-26 14:14:31Z: Job test completed with result: Failed
```

We have to support mounted dirs (typically specified with that flag) for this to be merged. Any thoughts are appreciated

Also looks like `/actions-runner` is involved. I have a PoC ill push up here to see what you think shortly

myoung34:
@marcus-bcl I made a PR to yours [here](https://github.com/marcus-bcl/docker-github-actions-runner/pull/1)

If thats merged the changes will propagate back here so you get credit for the PR

tested [here](https://github.com/OctoKode/test1/actions/runs/2740323802)

```
2022-07-26 14:40:48Z: Listening for Jobs
2022-07-26 14:41:48Z: Running job: test
2022-07-26 14:41:55Z: Job test completed with result: Succeeded
```

marcus-bcl:
I had the same issue initially with the default `/_work` dir - hence the `chown` step in the Dockerfile.

You need to ensure user 1001 in the container has write permissions on the host volume.  I believe this is a common issue with Docker - see https://github.com/moby/moby/issues/2259

To test it, you can do:
```shell
mkdir -p bar
sudo chown 1001 bar
docker run -it -e RUNNER_WORKDIR=/foo -v 'bar:/foo' myoung34/github-actions-runner
```

Alternatively, you can still revert to the old behaviour by running the container as root:
```shell
mkdir -p bar
docker run -it -u root -e RUNNER_WORKDIR=/foo -v 'bar:/foo' myoung34/github-actions-runner
```

Unfortunately your suggestion would mean the container initially has to run as root, which still prevents it from being deployed to restricted environments (e.g. Kubernetes clusters where runAsNonRoot is enforced).

myoung34:
> You need to ensure user 1001 in the container has write permissions on the host volume. I believe this is a common issue with Docker - see https://github.com/moby/moby/issues/2259

the issue with moby will not cross over well to containerd/crio which is where my concern lies

Id like to avoid having nuances per CRI and rely instead on `gosu` as I think it will cross over better across the board (k8s using contanerd natively going forward, etc)

marcus-bcl:
Another option would be to run the entrypoint as root but the cmd as runner, by switching the order in the Dockerfile.  I've added another PR to show how that would work, here: https://github.com/marcus-bcl/docker-github-actions-runner/pull/2

It's your call whichever way you think is best.  My use case is to deploy this to a restricted environment so running as root at all likely won't work for me, but I understand you need to consider other use cases too.

myoung34:
> Another option would be to run the entrypoint as root but the cmd as runner, by switching the order in the Dockerfile. I've added another PR to show how that would work, here: [marcus-bcl#2](https://github.com/marcus-bcl/docker-github-actions-runner/pull/2)
> 
> It's your call whichever way you think is best. My use case is to deploy this to a restricted environment so running as root at all likely won't work for me, but I understand you need to consider other use cases too.

My goal is to support a large variety of environments. My PR to yours will make this work as expected and indeed run as non root (process albeit not the container). You can specify `-u runner` after this is merged if you can maintain the moby nuance and this should let you run the entire container as non root so everyone wins

You can always fork beyond that but if `USER` causes dynamic directories to not behave then there's a larger set of users to consider. I do still think this is a good pr in the grand scheme if you're willing to merge mine into yours I'll merge this

marcus-bcl:
Yep you're totally right.  Hadn't considered that I could do the reverse and set the user to `runner` with the default working directory, happy to merge this 👍

Uzlopak:
Just a request: If you merge this and release, could you please add information to the release notes regarding the folder access stuff?

myoung34:
@marcus-bcl Im testing three scenarios now:

* run with `-u runner` with the default dir (any dynamic dir beyond that is on the user and nuances in the CRI)
* run with defaults with default dir and ensure the `runner` user works on the dynamic dir
* run with defaults with dynamic dirs and ensure the `runner` user works on the dynamic dir

@Uzlopak Ill update the docs for sure as well as a wiki entry

myoung34:
@Uzlopak @marcus-bcl the changes are being built

Added docs [here](https://github.com/myoung34/docker-github-actions-runner/wiki/Usage#non-root-runners)

nmalaguti:
> the changes are being built

@myoung34 It looks like https://github.com/myoung34/docker-github-actions-runner/actions/runs/2741219359 was cancelled and needs to be run to push the latest versions (only the base versions have been updated so far).

myoung34:
@nmalaguti youre correct, I wanted the bases to be built first and got distracted by work. its building now.

PR #234, opened 2022-08-29
Check to see if not running as root

### LINKED ISSUE #233: entrypoint.sh exits when RUN_AS_ROOT unset
I recently had an issue starting the container, the RUN_AS_ROOT env var was unset, and I got this message in the logs:

```
ERROR: RUN_AS_ROOT env var is set to true but the user has been overriden and is not running as root
```

But we are running as root, so this message is confounding... I managed to work around it by setting the env var RUN_AS_ROOT to an empty string.

Rumbles:
Looking at this bit of code it was checking to see if running as root, but the message says not running as root, I think the logic is backwards.

I also ran shellcheck on the file and it made some comments about the logic use generally, but I don't want to tinker too much:

```
In entrypoint.sh line 136:
  [[ $(id -u) -ne 0 ]] && ( "$@" ) || ( echo "ERROR: RUN_AS_ROOT env var is set to true but the user has been overriden and is not running as root"; exit 1 )
                       ^-- SC2015 (info): Note that A && B || C is not if-then-else. C may run when A is true.


In entrypoint.sh line 138:
  [[ $(id -u) -eq 0 ]] && ( /usr/sbin/gosu runner "$@" ) || ( "$@" )
                       ^-- SC2015 (info): Note that A && B || C is not if-then-else. C may run when A is true.
```

Closes #233

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -94,7 +94,7 @@ configure_runner() {
       ${_AUTO_UPDATE}
 
   [[ ! -d "${_RUNNER_WORKDIR}" ]] && mkdir "${_RUNNER_WORKDIR}"
-  
+
   [[ $(id -u) -eq 0 ]] && /usr/bin/chown -R runner ${_RUNNER_WORKDIR} /opt/hostedtoolcache/ /actions-runner || :
 }
 
@@ -133,7 +133,10 @@ fi
 
 
 if [[ ${_RUN_AS_ROOT} == "true" ]]; then
-  [[ $(id -u) -eq 0 ]] && ( "$@" ) || ( echo "ERROR: RUN_AS_ROOT env var is set to true but the user has been overriden and is not running as root"; exit 1 )
+  if [[ $(id -u) -ne 0 ]]; then
+    echo "ERROR: RUN_AS_ROOT env var is set to true but the user has been overridden and is not running as root. Currently running as $(whoami)"
+    exit 1
+  fi
 else
   [[ $(id -u) -eq 0 ]] && ( /usr/sbin/gosu runner "$@" ) || ( "$@" )
 fi
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -139,7 +139,6 @@ if [[ ${_RUN_AS_ROOT} == "true" ]]; then
     echo "ERROR: RUN_AS_ROOT env var is set to true but the user has been overridden and is not running as root, but UID '$(id -u)'"
     exit 1
   fi
-    echo "ERROR: RUN_AS_ROOT env var is set to true but the user has been overridden and is not running as root. Currently running as $(whoami)"
     exit 1
   fi
 else
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -140,5 +140,9 @@ if [[ ${_RUN_AS_ROOT} == "true" ]]; then
     exit 1
   fi
 else
-  [[ $(id -u) -eq 0 ]] && ( /usr/sbin/gosu runner "$@" ) || ( "$@" )
+  if [[ $(id -u) -eq 0 ]]; then
+    /usr/sbin/gosu runner "$@"
+  else
+    "$@"
+  fi
 fi
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -139,7 +139,6 @@ if [[ ${_RUN_AS_ROOT} == "true" ]]; then
     echo "ERROR: RUN_AS_ROOT env var is set to true but the user has been overridden and is not running as root, but UID '$(id -u)'"
     exit 1
   fi
-  fi
 else
   [[ $(id -u) -eq 0 ]] && ( /usr/sbin/gosu runner "$@" ) || ( "$@" )
 fi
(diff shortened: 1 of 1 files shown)


entrypoint.sh
```diff
@@ -133,7 +133,7 @@ fi
 
 
 if [[ ${_RUN_AS_ROOT} == "true" ]]; then
-  [[ $(id -u) -eq 0 ]] && ( "$@" ) || ( echo "ERROR: RUN_AS_ROOT env var is set to true but the user has been overriden and is not running as root"; exit 1 )
```
myoung34, line 136: what this translates to:

```
if _RUN_AS_ROOT is true: (This defaults to true, the only way not be true is to give it an explicit value)
    if id -u is 0 (root):
      run CMD
    else
      error and exit 1
```

This is correct. It means: If run as root set and true AND root: run it. else error
The only way this can fail is if RUN_AS_ROOT is anything other than `true` specifically and id -u is not 0. this would mean that the user running it is not in fact root

I would build your own version of this that spits out what `id -u` is here as well as the value of `_RUN_AS_ROOT` because my guess is that its `true` and not root

if RUN_AS_ROOT is anything other than specifically `true` you wouldnt end up here. 

> I managed to work around it by setting the env var RUN_AS_ROOT to an empty string
this would cause it to default to `true` no matter what

```
✗ export RUN_AS_ROOT=""
✗ _RUN_AS_ROOT=${RUN_AS_ROOT:="true"}
✗ echo $_RUN_AS_ROOT
true
```

Rumbles: In my case the user is root, I hadn't set RUN_AS_ROOT, but I hit this error, maybe because of https://github.com/koalaman/shellcheck/wiki/SC2015 ?

myoung34: Shell check may flag it but that's not the case here. I can't reproduce this

Rumbles: I did some more testing this morning, I made the following change to a running container, in entrypoint.sh:

```
  [[ $(id -u) -eq 0 ]] && ( "$@" ) || ( echo "ERROR: RUN_AS_ROOT env var is set to true but the user has been overriden and is not running as root"; exit 1 )
```

To become:

```
  [[ $(id -u) -eq 0 ]] && ( "$@" ) || ( echo "ERROR: RUN_AS_ROOT env var is set to true but the user has been overriden and is not running as root.  Current user configured as $(whoami)"; exit 1 )
```

This change just prints out the username of the current user, then I started the container with:

```
/usr/bin/docker run \
  --rm \
  --name="github-runner-${REPO}-$1" \
  --privileged \
  --dns 192.168.200.1 \
  --network=host \
  -e REPO_URL="https://github.com/intenthq/${REPO}" \
  -e ACCESS_TOKEN="${PAT_TOKEN}" \
  -e RUNNER_NAME="teamcity-${REPO}-$1" \
  -e RUNNER_WORKDIR="/tmp/github-runner/work_${REPO}-$1" \
  -u root \
  -v /var/run/docker.sock:/var/run/docker.sock \
  -v /tmp/github-runner:/tmp/github-runner \
  -v /dev/log:/dev/log \
  -e LABELS="${LABELS}" \
   rumbles/github-runner:testing
```

And I got this in the logs:


```
Aug 30 07:57:37 build-1 github-runner.sh[2128657]: ERROR: RUN_AS_ROOT env var is set to true but the user has been overriden and is not running as root. Current user configured as root
```

I think with the current logic, if either of these fail:

```
 [[ $(id -u) -eq 0 ]] && ( "$@" )
```

The error state can run, which I don't think is what is wanted, it should only run if RUN_AS_ROOT is true but the user is not root.

My initial PR had made the wrong assumption, I will update it

entrypoint.sh
```diff
@@ -133,7 +133,10 @@ fi
...
+  if [[ $(id -u) -ne 0 ]]; then
```
myoung34: youve lost the part where if RUN_AS_ROOT is true and id -u is 0 that it still needs to run $@

```suggestion
  if [[ $(id -u) -eq 0 ]]; then
    "$@"
  else
    echo "ERROR: RUN_AS_ROOT env var is set to true but the user has been overridden and is not running as root, but UID '$(id -u)'"
    exit 1
  fi
```

entrypoint.sh
```diff
@@ -133,7 +133,10 @@ fi
...
+    echo "ERROR: RUN_AS_ROOT env var is set to true but the user has been overridden and is not running as root. Currently running as $(whoami)"
```
```suggestion
```

entrypoint.sh
```diff
@@ -133,7 +133,10 @@ fi
...
+    exit 1
```
myoung34, line 140:
```suggestion
```

entrypoint.sh
```diff
@@ -133,7 +133,10 @@ fi
...
+  fi
```
myoung34, line 141:
```suggestion
```

entrypoint.sh
```diff
@@ -133,7 +133,10 @@ fi
...
 else
   [[ $(id -u) -eq 0 ]] && ( /usr/sbin/gosu runner "$@" ) || ( "$@" )
```
myoung34: if you want the fix above you should likely port it here too

myoung34 (changes requested):

entrypoint.sh
Rumbles: Agreed, I have updated that section as well, thanks for the feedback, sorry for the slow reply

PR #238, opened 2022-09-06
Change gh cli install method to github releases

### LINKED ISSUE #236: GitHub CLI GPG key expired
Hi,

https://github.com/cli/cli/issues/6175

The key that has been used to sign the GitHub CLI package has expired, and will cause errors on subsequent apt updates and apt installs unless apt is told otherwise.

It seems to me that they won't be renewing/adding new keys to sign that package, and will instead rely on workflows pulling directly from the release page (on GitHub)

This means that these lines must be scrapped https://github.com/myoung34/docker-github-actions-runner/blob/master/Dockerfile.base#L73-L77
Instead the build for this image can do a curl/wget to get the current stable release from https://github.com/cli/cli/releases/ and put it somewhere in PATH.

A temporary workaround for users that are hitting this issue in their workflows (if you are installing packages using apt):
`(sudo) rm /etc/apt/sources.list.d/github-cli.list`

nmalaguti:
The apt repo for the gh cli is broken due to an expired GPG key: https://github.com/cli/cli/issues/6175

They want to retire the debian packaging altogether https://github.com/cli/cli/issues/5941 and this is the recommended "long term" solution https://github.com/cli/cli/issues/6175#issuecomment-1235984381.

Adapted from https://github.com/cli/cli/issues/6175#issuecomment-1236264607

Fixes #236

### DIFF
--- a/Dockerfile.base
+++ b/Dockerfile.base
@@ -46,6 +46,8 @@ RUN echo en_US.UTF-8 UTF-8 >> /etc/locale.gen \
     nodejs \
     rsync \
     gosu \
+  && DPKG_ARCH="$(dpkg --print-architecture)" \
+  && LSB_RELEASE_CODENAME="$(lsb_release --codename | cut -f2)" \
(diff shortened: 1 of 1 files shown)


myoung34:
Nice fix!

myoung34:
Ironically i merged this _after_ it apparently got fixed upstream SMH

PR #250, opened 2022-09-30
Fix ephemeral support

### LINKED ISSUE #249: Ephemeral mode not working since 08608da938e5f565c4a9f38a793a2d29ad5db1bd
In https://github.com/myoung34/docker-github-actions-runner/commit/08608da938e5f565c4a9f38a793a2d29ad5db1bd

double-quotes got added around `_EPHEMERAL` and `_AUTO_UPDATE`.

This has the side effect of triggering config with the following arguments when EPHEMERAL is set, but not AUTO_UPDATE:

```
./config --url https://github.com/org/repo--token  ABCD --name  somenames --work /tmp/work --labels label1,label2 --runnergroup Default --unattended --replace --ephemeral ''
```

the reason is illustrated in https://gist.github.com/chantra/6a7c526f45f833da7eb7d08b053757d8

Given how actions runner are handling command line arguments: https://github.com/actions/runner/blob/3a1c89715cccc9ec2cf55c0903a8466de839d7fe/src/Runner.Common/CommandLineParser.cs#L88

`''` become a value for `--ephemeral`, which likely is interpreted as "false".

I have tested by removing the double-quotes around `_AUTO_UPDATE` and the container was started as ephemeral again.

I suppose the solution here is either:
- remove the double-quotes and tell shellcheck this is fine
- use an array where optional arguments gets added when needed and pass this to config.

chantra:
The quotes added in 08608da938e5f565c4a9f38a793a2d29ad5db1bd cause `ephemeral` to evaluate as false when `DISABLE_AUTO_UPDATE` was not set.

This is explained in details in #249. This diff changes the logic to use an array where extra parameters are being added.
When calling config.sh, the array is expanded.

Testing:
../self_hosted_runner.env contains
```
EPHEMERAL=true
```

Before: https://gist.github.com/chantra/443dfaa60d964d938bcb1c020781ca00 we can see multiple jobs got scheduled on the runner

After: https://gist.github.com/chantra/061dc471c63964c01a2e3733a53bd3ec we can see that when the job finished, the container shutdown.

After with DISABLE_AUTO_UPDATE set: https://gist.github.com/chantra/df0102c7340b8f6bfd458ec56d9fa29c

Also adding `set -x` before the call to `config.sh` revealed the arguments were correctly added.

Fixes #249

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -27,7 +27,6 @@ _LABELS=${LABELS:-default}
 _RUNNER_GROUP=${RUNNER_GROUP:-Default}
 _GITHUB_HOST=${GITHUB_HOST:="github.com"}
 _RUN_AS_ROOT=${RUN_AS_ROOT:="true"}
-_EPHEMERAL=""
 
 # ensure backwards compatibility
 if [[ -z $RUNNER_SCOPE ]]; then
@@ -62,6 +61,7 @@ case ${RUNNER_SCOPE} in
 esac
 
 configure_runner() {
+  ARGS=()
   if [[ -n "${ACCESS_TOKEN}" ]]; then
     echo "Obtaining the token of the runner"
     _TOKEN=$(ACCESS_TOKEN="${ACCESS_TOKEN}" bash /token.sh)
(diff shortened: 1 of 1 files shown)


myoung34:
Thanks for the fix!

PR #256, opened 2022-10-31
Add clang and libclang1 to base image

klaus993:
This is needed for several language's libraries, I don't think it makes any harm and it's useful to a bunch of people, including me 😛 

I was wondering if you'd include it in your base image, otherwise I'll have to work with a fork 😄 

Thanks in advance.

### DIFF
--- a/Dockerfile.base
+++ b/Dockerfile.base
@@ -18,6 +18,8 @@ RUN echo en_US.UTF-8 UTF-8 >> /etc/locale.gen \
   && apt-key adv --keyserver keyserver.ubuntu.com --recv-keys ${GIT_CORE_PPA_KEY} \
   && apt-get update \
   && apt-get install -y --no-install-recommends \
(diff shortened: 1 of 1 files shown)


klaus993:
cc @myoung34

JulianGro:
I didn't check very deeply, but according to packages.debian.org this would add at least 80 MiB to the image. 
For reference, I don't use clang in this scenario.

klaus993:
Ah okay, if being as lightweight as possible is a concern, then just close my PR 😄

myoung34:
Ill leave this open as I dont have a personal opinion other than 80 is quite a bit

If others seem to agree we can def add it, but it might be worthwhile to add another build on top of each final output for larger images that contain a lot more packages as an optional pull vs mandatory

myoung34:
Closing for more info/context in #259

PR #264, opened 2022-12-06
Allow to put APP_PRIVATE_KEY in Docker env file. Introduced APP_LOGIN…

Abramo-Bagnara:
… to fix access token retrieval in case of multiple GitHub's App subscriber.

Before this patch the token retrieval failed because jq extracted more than URL.

### DIFF
--- a/app_token.sh
+++ b/app_token.sh
@@ -4,6 +4,7 @@
 # Environment variable that need to be set up:
 # * APP_ID, the GitHub's app ID
 # * APP_PRIVATE_KEY, the content of GitHub app's private key in PEM format.
+# * APP_LOGIN, the login name used to install GitHub's app
 #
 # https://github.com/orgs/community/discussions/24743#discussioncomment-3245300
 #
@@ -66,7 +67,7 @@ rs256_sign() {
 request_access_token() {
     jwt_payload=$(build_jwt_payload)
     encoded_jwt_parts=$(base64url <<<"${JWT_JOSE_HEADER}").$(base64url <<<"${jwt_payload}")
-    encoded_mac=$(echo -n "$encoded_jwt_parts" | rs256_sign "${APP_PRIVATE_KEY}" | base64url)
+    encoded_mac=$(echo -n "${encoded_jwt_parts}" | rs256_sign "${APP_PRIVATE_KEY}" | base64url)
     generated_jwt="${encoded_jwt_parts}.${encoded_mac}"
 
     auth_header="Authorization: Bearer ${generated_jwt}"
@@ -74,9 +75,9 @@ request_access_token() {
     app_installations_response=$(curl -sX GET \
         -H "${auth_header}" \
         -H "${API_HEADER}" \
-        ${APP_INSTALLATIONS_URI} \
+        "${APP_INSTALLATIONS_URI}" \
     )
-    access_token_url=$(echo "$app_installations_response" | jq --raw-output '.[] | select (.app_id  == '"${APP_ID}"') .access_tokens_url')
+    access_token_url=$(echo "${app_installations_response}" | jq --raw-output '.[] | select (.account.login == "'"${APP_LOGIN}"'" and .app_id  == '"${APP_ID}"') .access_tokens_url')
     curl -sX POST \
         -H "${CONTENT_LENGTH_HEADER}" \
         -H "${auth_header}" \
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -47,6 +47,9 @@ case ${RUNNER_SCOPE} in
     [[ -z ${ORG_NAME} ]] && ( echo "ORG_NAME required for org runners"; exit 1 )
     _SHORT_URL="https://${_GITHUB_HOST}/${ORG_NAME}"
     RUNNER_SCOPE="org"
+    if [[ -n "${APP_ID}" ]] && [[ -z "${APP_LOGIN}" ]]; then
+      APP_LOGIN=${ORG_NAME}
+    fi
     ;;
 
   ent*)
@@ -59,6 +62,10 @@ case ${RUNNER_SCOPE} in
     [[ -z ${REPO_URL} ]] && ( echo "REPO_URL required for repo runners"; exit 1 )
     _SHORT_URL=${REPO_URL}
     RUNNER_SCOPE="repo"
+    if [[ -n "${APP_ID}" ]] && [[ -z "${APP_LOGIN}" ]]; then
+      APP_LOGIN=${REPO_URL%/*}
+      APP_LOGIN=${APP_LOGIN##*/}
+    fi
     ;;
 esac
 
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -2,7 +2,7 @@
 # shellcheck shell=bash
 
 export RUNNER_ALLOW_RUNASROOT=1
-export PATH=$PATH:/actions-runner
+export PATH=${PATH}:/actions-runner
 
 # Un-export these, so that they must be passed explicitly to the environment of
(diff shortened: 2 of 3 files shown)


myoung34:
I think id like to make this optional, not mandatory as this PR places it.
There will be many that dont have or need the APP_LOGIN, so adding a code path for it as an optional would be ideal

Abramo-Bagnara:
> I think id like to make this optional, not mandatory as this PR places it. There will be many that dont have or need the APP_LOGIN, so adding a code path for it as an optional would be ideal

I'm not sure to understand the part "don't have the APP_LOGIN": every installed GitHub's App  have necessarily a login associated.

Note also that to not add an APP_LOGIN when multiple login exists for a given app leads to obscure failures not easy to interpret.

IMHO to ask to specify the wanted login to be searched for the token is the safest alternative.

myoung34:
> I'm not sure to understand the part "don't have the APP_LOGIN": every installed GitHub's App have necessarily a login associated.

The code currently assumes that there's only one app and will pull it down via 

```
    access_token_url=$(echo "$app_installations_response" | jq --raw-output '.[] | select (.app_id  == '"${APP_ID}"') .access_tokens_url')
```

which works. This change introduces `APP_LOGIN` whose only job is an addition select in the jq

```
    access_token_url=$(echo "${app_installations_response}" | jq --raw-output '.[] | select (.account.login == "'"${APP_LOGIN}"'" and .app_id  == '"${APP_ID}"') .access_tokens_url')
```

My point here is that it's not 100% necesary as-is, so making it optional is ideal as its introducing a breaking change of sorts

Abramo-Bagnara:
Perhaps I have not been clear: current jq extraction does not work as it retrieve more than one access_tokens_url when there are multiple login on the app.
If you want it somewhat "kind of works" (i.e. extract only the first, that perhaps is not the desired one) you have to change the jq selection to use only the first found, but in this case it seems pointless to me to not do the right thing, namely to search the login that user wants.
If you really are contrary to specify that, a default might be perhaps either extracted either from REPO_URL (the part before the last) or copied from ORG_NAME.

myoung34:
As long as theres a safe default that works as-is then I think it's fine, I just don't wish to require a new for those are arent currently using it and will find things failing

entrypoint.sh
```diff
@@ -59,20 +62,26 @@ case ${RUNNER_SCOPE} in
...
+      APP_LOGIN=${REPO_URL%/*}
+      APP_LOGIN=${APP_LOGIN##*/}
+    fi
     ;;
 esac
 
 configure_runner() {
   ARGS=()
-  if [[ -n "${APP_ID}" ]] && [[ -n "${APP_PRIVATE_KEY}" ]]; then
+  if [[ -n "${APP_ID}" ]] && [[ -n "${APP_PRIVATE_KEY}" ]] && [[ -n "${APP_LOGIN}" ]]; then
```
myoung34, line 74: no need for `&& [[ -n "${APP_LOGIN}" ]]` anymore since it has a default and will be set if APP_ID is there (Line 50 and 65)

entrypoint.sh
```diff
@@ -59,20 +62,26 @@ case ${RUNNER_SCOPE} in
...
-      echo "ERROR: ACCESS_TOKEN or RUNNER_TOKEN provided but are mutually exclusive with APP_ID and APP_PRIVATE_KEY." >&2
+      echo "ERROR: ACCESS_TOKEN or RUNNER_TOKEN provided but are mutually exclusive with APP_ID, APP_PRIVATE_KEY and APP_LOGIN." >&2
       exit 1
     fi
-    echo "Obtaining access token for app_id ${APP_ID}"
-    ACCESS_TOKEN=$(APP_ID="${APP_ID}" APP_PRIVATE_KEY="${APP_PRIVATE_KEY}" bash /app_token.sh)
-  elif [[ -n "${APP_ID}" ]] || [[ -n "${APP_PRIVATE_KEY}" ]]; then
-    echo "ERROR: Both APP_ID and APP_PRIVATE_KEY must be specified." >&2
+    echo "Obtaining access token for app_id ${APP_ID} and login ${APP_LOGIN}"
+    nl="
```
myoung34, line 80: Not sure I understand what `nl` is for here

entrypoint.sh
```diff
@@ -59,20 +62,26 @@ case ${RUNNER_SCOPE} in
...
+"
+    ACCESS_TOKEN=$(APP_ID="${APP_ID}" APP_PRIVATE_KEY="${APP_PRIVATE_KEY//\\n/${nl}}" APP_LOGIN="${APP_LOGIN}" bash /app_token.sh)
+  elif [[ -n "${APP_ID}" ]] || [[ -n "${APP_PRIVATE_KEY}" ]] || [[ -n "${APP_LOGIN}" ]]; then
```
myoung34, line 83: `|| [[ -n "${APP_LOGIN}"` isnt needed here either since it has a default value

myoung34 (changes requested):

myoung34:
Also make sure to add some metadata to the README for APP_LOGIN please

entrypoint.sh
Abramo-Bagnara: It is needed to allow to store APP_PRIVATE_KEY in .env file once replaced newlines with \n.
These are converted back to newline here.

entrypoint.sh
Abramo-Bagnara: There is not a default value when scope is enterprise.

entrypoint.sh
myoung34: This seems like an implementation detail and shouldnt be logic in the entrypoint, but Im still not sure I follow why wed approach it this way. I dont think I want to add that logic here to handle .env specific parsing if others havent had this issue

entrypoint.sh
myoung34: We should default to ORG_NAME in all cases so that it doesn't behave differently for others not using APP_LOGIN currently with success

myoung34 (changes requested):

entrypoint.sh
Abramo-Bagnara: Everybody has this issue when using APP_PRIVATE_KEY: it is a well known problem for Docker .env files that need to be solved by scripts inside the docker.

entrypoint.sh
Abramo-Bagnara: Doing that you would hide a sane error ("please specify APP_LOGIN") with an obscure failure after jq query when multiple urls are returned instead of the single one expected.

app_token.sh
```diff
@@ -66,17 +67,17 @@ rs256_sign() {
...
     auth_header="Authorization: Bearer ${generated_jwt}"
 
     app_installations_response=$(curl -sX GET \
         -H "${auth_header}" \
         -H "${API_HEADER}" \
-        ${APP_INSTALLATIONS_URI} \
+        "${APP_INSTALLATIONS_URI}" \
     )
-    access_token_url=$(echo "$app_installations_response" | jq --raw-output '.[] | select (.app_id  == '"${APP_ID}"') .access_tokens_url')
+    access_token_url=$(echo "${app_installations_response}" | jq --raw-output '.[] | select (.account.login == "'"${APP_LOGIN}"'" and .app_id  == '"${APP_ID}"') .access_tokens_url')
```
myoung34, line 80: Tested with

```
✗ export APP_LOGIN="foo"
✗ export APP_ID=1
✗ echo '[{"account": {"login": "foo"}, "app_id": 1, "access_tokens_url": "wut"}]'  | jq --raw-output '.[] | select (.account.login == "'"${APP_LOGIN}"'" and .app_id  == '"${APP_ID}"') .access_tokens_url'

wut
```

myoung34 (approved):

PR #271, opened 2023-01-16
Added using docker hostname instead manual generated random

iamironz:
Hi! 
To be able to use docker's `--scale` option in a large number of containers while debugging particular instances I found it way more convenient to use the container name to reflect the name in the GitHub repository's admin instead of randomly generated IDs. 
Also using the host machine volume directory is better with the docker container id name. 

`cat /etc/hostname` - will output docker generated id which is easily visible in `docker ps -a` list.
 `_RUNNER_NAME` as the work dir name is convenient when you need to visit the runner directory without having to navigate internal docker volumes.

Thank you for your effort.


myoung34 (approved):

myoung34:
Tested in docker proper and k3s (containerd)
Looks good, thanks!

PR #276, opened 2023-01-30
Bugfix: /etc/hostname is not guaranteed, sane default to random string

### LINKED ISSUE #275: `_RUNNER_NAME` does not given properly when runs container on Cloud Run and ECS
## Description
The runner name is not given randomly when using v2.301.1 with Cloud Run and ECS on Fargate.

I'm running self-hosted runner on Google Cloud Cloud Run Job and AWS ECS on Fargate with my custom image based on this repository. Since the release of `myoung34/github-runner:2.301.1` my runners could not launch concurrently.
The reason is that all runners try to use the same name, but GitHub does not accept the same runner name as you know.

- Cloud Run job: `${RUNNER_NAME_PREFIX}-worker`. 
- ECS on Fargate: `{RUNNER_NAME_PREFIX}-` 

https://github.com/myoung34/docker-github-actions-runner/pull/271 changed to get random string from /etc/hostname instead of self generated random string, but some container platform does not provide random string.
In my expectation, Cloud Run Job provide fix hostname `worker` and ECS on Fargate provide empty string or maybe /etc/hostname does not exist.

## Solutions I would expect
Could you add new option that user can choose previous behavior that generates random string as `_RUNNER_NAME` ? Or please revert to https://github.com/myoung34/docker-github-actions-runner/pull/271.

myoung34:
Resolves #275

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,8 @@ These containers are built via Github actions that [copy the dockerfile](https:/
 | --- | --- |
 | `RUN_AS_ROOT` | Boolean to run as root. If `true`: will run as root. If `True` and the user is overridden it will error. If any other value it will run as the `runner` user and allow an optional override. Default is `true` |
 | `RUNNER_NAME` | The name of the runner to use. Supercedes (overrides) `RUNNER_NAME_PREFIX` |
-| `RUNNER_NAME_PREFIX` | A prefix for runner name (followed by docker container id if /etc/hostname if exists, else a random 13 digit string). You must not also provide `RUNNER_NAME`. Defaults to `github-runner`                                                                                                                                                              |
+| `RUNNER_NAME_PREFIX` | A prefix for runner name (See `RANDOM_RUNNER_SUFFIX` for how the full name is generated). Note: will be overridden by `RUNNER_NAME` if provided. Defaults to `github-runner` |
+| `RANDOM_RUNNER_SUFFIX` | Boolean to use a randomized runner name suffix (preceeded by `RUNNER_NAME_PREFIX`). Will use a 13 character random string by default. If set to a value other than true it will attempt to use the contents of `/etc/hostname` or fall back to a random string if the file does not exist or is empty. Note: will be overridden by `RUNNER_NAME` if provided. Defaults to `true`. |
 | `ACCESS_TOKEN` | A [github PAT](https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token) to use to generate `RUNNER_TOKEN` dynamically at container start. Not using this requires a valid `RUNNER_TOKEN` |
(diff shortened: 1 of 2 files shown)


Kesin11:
I found ECS on fargate has /etc/hostname but it's empty, so this patch doesn't fix #275 unfortunately.

I'm sorry https://github.com/myoung34/docker-github-actions-runner/issues/275#issuecomment-1408964511 was wrong. I checked /etc/hostname file exists using overwrite container CMD that ECS provided but it's not the right way.
I tried to attach container that built from this patch using [ECS Exec](https://docs.aws.amazon.com/AmazonECS/latest/userguide/ecs-exec.html) and found ECS on fargate has /etc/hostname but it's empty. Here is a log.

```
$ aws ecs execute-command --region MY_REGION --cluster MY_CLSTER --task cc296bf1bdc3463fbddca42796d4b66a --interactive --command "/bin/sh"

The Session Manager plugin was installed successfully. Use the AWS CLI to start a session.


Starting session with SessionId: ecs-execute-command-0cca150f7db8f5210
# ls /etc/hostname
/etc/hostname
# cat /etc/hostname

```

myoung34:
@Kesin11 Try now
I added a check to make sure it exists but also is not empty

Kesin11:
I tried on Cloud Run and ECS on Fargate.

When runs on Cloud Run with set `RANDOM_RUNNER_SUFFIX` to true, runner name suffix is now a random string. It works perfectly!

Unfortunately, when runs on ECS with set `RANDOM_RUNNER_SUFFIX` to false, runner name suffix is still empty.
Here is runner log `RANDOM_RUNNER_SUFFIX is false. /etc/hostname exists and has content. Setting runner name to ecs-`

I looked again at what was going on in the ECS container. As a result, /etc/hostname does not empty but it has only one empty character!

```
Starting session with SessionId: ecs-execute-command-080c21d7fec1b1248
# stat --printf="%s" /etc/hostname
1#
# cat /etc/hostname

# wc -l /etc/hostname
1 /etc/hostname
#
# cat -A /etc/hostname
$
```

Is this a newline character?

Kesin11:
@myoung34 
To handle with the strange behavior of ECS, we may need to check that /etc/hostname is less than one character, or check it's content using regular expression.

myoung34:
@Kesin11 you can use the new environment variable in ecs to skip the hostname behavior altogether 

Simply don't set it or be explicit and set it to true and it won't try to use that file regardless of size

Kesin11:
@myoung34 
I agree. In my case it is enough if a random string without respecting /etc/hostname. I'll set `RANDOM_RUNNER_SUFFIX` to true or false for each container platform.

myoung34:
Thanks for verifying @Kesin11 Ill merge as is

if you notice anything open a new Issue and we'll get you sorted. :tada:

PR #282, opened 2023-02-07
Make default workdir a subfolder under /_work

fabfuel:
Hi @myoung34,

first of all, thanks a lot for this great project!

I have currently the problem, that I can not share files between "native" GitHub actions, which run inside a container based on your image and other actions wich run in their own Docker containers. 

The problem is, that there is no common/shared workdir. Currently the default workdir might be `/_work-oERTjrGWXtKcd` and in another worker `/_work-jHahu2ueAudhe`, etc. 

This PR makes the default workdir a sub directory under `/_work/`, for example `/_work/oERTjrGWXtKcd`.
This way, you can make `/_work/` a volume for all runner containers and share files between actions.

Best
Fabian

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -41,7 +41,7 @@ if [[ ${RANDOM_RUNNER_SUFFIX} != "true" ]]; then
   fi
 fi
 
-_RUNNER_WORKDIR=${RUNNER_WORKDIR:-/_work-${_RUNNER_NAME}}
+_RUNNER_WORKDIR=${RUNNER_WORKDIR:-/_work/${_RUNNER_NAME}}
 _LABELS=${LABELS:-default}
 _RUNNER_GROUP=${RUNNER_GROUP:-Default}
 _GITHUB_HOST=${GITHUB_HOST:="github.com"}


fabfuel:
> You can set the RUNNER_WORKDIR env var to /_work/ though correct?

Yes, but then all runners use the same workdir and If I mount that from the Host, all runners will write into the same dir. It's very likely, that parallel jobs will overwrite each others files, which I would like to avoid.

Another solution could be to only provide a `RUNNER_WORKDIR_PREFIX`, but that the entry point still adds the random part to it.

fabfuel:
FYI: I'm now using a fork with exactly that change in our CI env and it works fine. 
I now can share files between native jobs and those jobs which run inside a separate Docker container.

Uzlopak:
Can you elaborate this please?
Is that related to some docker in docker restrictions?

Uzlopak:
Actually you can configure the RUNNER_NAME and RUNNER_WORKDIR. So there is no real need for this PR

fabfuel:
Yes, exactly:
If you have a step, which runs natively, it is running inside the runner container (this project). If files are written during that step (for example code coverage results), they end up being stored within the container.

A subsequent step, which also runs natively, can access this file without problem.

But if a step is using its own Docker image, it runs in a separate container, files written in there are not available from the native processes. This can be solved by:
1. mounting the workdir in the runner container, and
2. mounting the workdir in the custom container (which happens by default)

By mounting `/_work/` in the runner, all files written, end up being stored on the Docker host, which means all native steps and all custom steps (running in separate containers) can access the files of a pipeline run.

Use-cases which need that functionality, that I stumbled upon so far:
- publishing build or test artifacts in GitHub Actions (e.g. Cypress videos)
- uploading code coverage files, which are generated during a job that should be uploaded to an external provider which provides their own GitHub Action (e.g. Codacy)
- downloading translation files from an external provider into the workdir, with the providers GitHub Action (e.g. Localazy)
- etc.

Best
Fabian

fabfuel:
> Actually you can configure the RUNNER_NAME and RUNNER_WORKDIR. So there is no real need for this PR

The runner name is only used, if I do NOT provide a workdir. 
So yes, there is a real need.

Uzlopak:
So you still have to mount the working directory into the custom container? How you solve that?

Uzlopak:
Well. You can specify RUNNER_NAME and RUNNER_WOKDIR and  I dont see why it should not work. Also your solution could break the separation of docker containers so I strongly recommend to not do it this way.

myoung34:
I dont see much risk here unless you can fault me:

* no change - you can set RUNNER_WORKDIR to something. 
* some change - you can set RUNNER_NAME and get `/_work/{SET NAME}` vs `/_work-{SET NAME}`
* some change - do nothing and get `/_work/{GENERATED RUNNER NAME}` vs `/_work-{GENERATED RUNNER NAME}`

did i miss a scenario?
I dont see any risk of compatibility breakage for `/_work/{SET NAME}` vs `/_work-{SET NAME}`

fabfuel:
Thanks for chiming in @myoung34!

I think this is a good summary, I also do not see any other risk of compatibility breakage, as the workdir is already randomly generated, if you do not specify `RUNNER_WORKDIR` or `RUNNER_NAME`.

If I understood the readme correctly, my change restores the behaviour mentioned here:
https://github.com/myoung34/docker-github-actions-runner/blob/master/README.md?plain=1#L70

> Runners on the same host should not share this directory

Currently this is not possible, by setting the `RUNNER_WORKDIR`, as then all runners would use the same workdir, so there needs to be randomness, which is currently skipped, by setting the workdir statically. Therefore a `RUNNER_WORKDIR_PREFIX` (as it exists for the name) would be another nice solution.

Best
Fabian

myoung34:
This will go out as `:latest` automatically. I don't retag old releases so whenever actions gets a new version it will be in those going forward. If you don't mind testing `:latest` when it goes out you can reopen an issue if it doesnt work as expected

fabfuel:
Awesome, thanks @myoung34! I will test in on Monday.

PR #284, opened 2023-02-28
feat: add Docker Dind support

haridarshan:
Recently I came accross an issue where Docker-based GitHub Action ([Super Linter](https://github.com/github/super-linter)) is not working because it requires container inside container and because of this I came across a solution provided by Docker itself [Docker in Docker](https://github.com/docker-library/docs/tree/master/docker#what-is-docker-in-docker).

I've tried to add [Docker:dind](https://hub.docker.com/layers/library/docker/dind/images/sha256-ed6220b0de0f309f0844cf8cf1a6b861e981fb7f5c28bec6acc97abc910bd0a8?context=explore) in this image so that Docker-based GitHub Action doesn't fail.

Please Note: this will require to run docker container with `--privileged` flag

### DIFF
--- a/Dockerfile.base
+++ b/Dockerfile.base
@@ -92,7 +92,7 @@ RUN echo en_US.UTF-8 UTF-8 >> /etc/locale.gen \
   && echo '%sudo ALL=(ALL) NOPASSWD: ALL' >> /etc/sudoers
 
 COPY modprobe.sh /usr/local/bin/modprobe
-COPY docker-entrypoint.sh /usr/local/bin/
+COPY docker-entrypoint.sh dockerd-entrypoint.sh /usr/local/bin/
 
 ENV DOCKER_TLS_CERTDIR=/certs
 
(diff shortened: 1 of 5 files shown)


myoung34:
This is documented as not currently possible [here](https://github.com/myoung34/docker-github-actions-runner#docker-support)

Also while I appreciate the effort, docker is already installed in [the base image](https://github.com/myoung34/docker-github-actions-runner/blob/master/Dockerfile.base#L66) all that's needed is for Github to allow DinD and pass it the docker socket via `-v`. The link you used to build up this PR is almost 6 years out of date


```
➜ docker pull myoung34/github-runner:latest >/dev/null 2>&1

➜ docker run -it --entrypoint /bin/bash -v /var/run/docker.sock:/var/run/docker.sock myoung34/github-runner:latest
root@75ac2fed1fa9:/actions-runner# docker ps
CONTAINER ID   IMAGE                           COMMAND       CREATED         STATUS        PORTS     NAMES
75ac2fed1fa9   myoung34/github-runner:latest   "/bin/bash"   2 seconds ago   Up 1 second             elegant_almeida

root@75ac2fed1fa9:/actions-runner#
```

Uzlopak:
FYI 

supporting DinD is on the roadmap of the github runner project planned for this quarter (Q1 / 2023).

PR #286, opened 2023-03-17
add whitespace around gh version number

### LINKED ISSUE #274: Make Dockerfile easier to patch
Hi. I'm using your "github actions runner in a docker". I patch the Dockerfile to run on linux/arm64 with ```ARG TARGETPLATFORM="linux/arm64"```. Unfortunately, that is the line just after the github runner version. So every time the github runner version changes the patch fails. 
Could you put a few lines before and after ``ARG GH_RUNNER_VERSION``, like in the attached file, 
[Dockerfile.txt](https://github.com/myoung34/docker-github-actions-runner/files/10465103/Dockerfile.txt), to keep the github runner version a bit separate from other variables? Thanks.

koendv:
This PR adds whitespace around the github version number in the Dockerfile, so a patch to the Dockerfile stays the same, even if the github version number changes. Closes #274

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,6 @@ RUN mkdir -p /opt/hostedtoolcache
 
 ARG GH_RUNNER_VERSION="2.303.0"
 
-
 ARG TARGETPLATFORM
 
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,12 @@ LABEL maintainer="myoung34@my.apsu.edu"
 ENV AGENT_TOOLSDIRECTORY=/opt/hostedtoolcache
 RUN mkdir -p /opt/hostedtoolcache
 
+
+# GH_RUNNER_VERSION
 ARG GH_RUNNER_VERSION="2.303.0"
+
+
+
 ARG TARGETPLATFORM
 
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,6 @@ RUN mkdir -p /opt/hostedtoolcache
 ARG GH_RUNNER_VERSION="2.303.0"
 
 
-
 ARG TARGETPLATFORM
 
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,6 @@ LABEL maintainer="myoung34@my.apsu.edu"
 ENV AGENT_TOOLSDIRECTORY=/opt/hostedtoolcache
 RUN mkdir -p /opt/hostedtoolcache
 
-# GH_RUNNER_VERSION
 ARG GH_RUNNER_VERSION="2.303.0"
 
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,6 @@ LABEL maintainer="myoung34@my.apsu.edu"
 ENV AGENT_TOOLSDIRECTORY=/opt/hostedtoolcache
 RUN mkdir -p /opt/hostedtoolcache
 
-
 # GH_RUNNER_VERSION
 ARG GH_RUNNER_VERSION="2.303.0"
 


myoung34:
Im very confused. How does this change anything externally? Im not sure how this makes anything easier to patch

What are you currently doing to patch? 

I honestly don't think I understand what "so a patch to the Dockerfile stays the same"

This doesn't change the fact that if my automation bumps the version there will be a diff to the Dockerfile.

Do you have any screenshots or snippets of what youre doing and how this resolves it?

koendv:
On Fri, 17 Mar 2023 10:24:18 -0700
myoung34 ***@***.***> wrote:

> Im very confused. How does this change anything externally? Im not
> sure how this makes anything easier to patch
> 
> What are you currently doing to patch?
> 

OK. This is the project:
https://github.com/koendv/arduino-ide-raspberrypi/
and this is the patch to your Dockerfile:
https://github.com/koendv/arduino-ide-raspberrypi/blob/main/docker-github-actions-runner.patch
I'd like to automate the build.
As you can see, I modify ARG TARGETPLATFORM. This is the line after GH_RUNNER_VERSION.
Because patch typically looks 3 lines before and 3 lines after what you wish to patch, each time GH_RUNNER_VERSION changes the patch fails.
Add three lines of whitespace before and after "ARG GH_RUNNER_VERSION=", and if you change TARGETPLATFORM the diff no longer contains GH_RUNNER_VERSION.

best,

koen

myoung34:
I see now. I tested this locally with only a single newline after `ARG GH_RUNNER_VERSION=...` and it worked fine. Ill suggest edits you can accept

Dockerfile
```diff
@@ -5,7 +5,12 @@ LABEL maintainer="myoung34@my.apsu.edu"
 ENV AGENT_TOOLSDIRECTORY=/opt/hostedtoolcache
 RUN mkdir -p /opt/hostedtoolcache
 
+
```
myoung34:
```suggestion
```

Dockerfile
```diff
@@ -5,7 +5,12 @@ LABEL maintainer="myoung34@my.apsu.edu"
...
+# GH_RUNNER_VERSION
```
myoung34:
```suggestion
```

Dockerfile
```diff
@@ -5,7 +5,12 @@ LABEL maintainer="myoung34@my.apsu.edu"
...
 ARG GH_RUNNER_VERSION="2.303.0"
+
+
```
myoung34:
```suggestion
```

Dockerfile
```diff
@@ -5,7 +5,12 @@ LABEL maintainer="myoung34@my.apsu.edu"
...
+
```
myoung34:
```suggestion
```

myoung34 (changes requested):

myoung34:
Its worth noting you can (and should) be using TARGETPLATFORM as a build param, but my suggestions solve it regardless

Dockerfile
myoung34: Theres only need for one newline to avoid the merge conflict

myoung34 (changes requested):

koendv:
The GH_VERSION_NUMBER still appears:

***@***.***:~/src/docker-github-actions-runner $ git diff
diff --git a/Dockerfile b/Dockerfile
index 7f72fdf..8b13e9a 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,7 @@ RUN mkdir -p /opt/hostedtoolcache
 ARG GH_RUNNER_VERSION="2.303.0"
 
 
-ARG TARGETPLATFORM
+ARG TARGETPLATFORM="linux/arm64"
 
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 



On Fri, 17 Mar 2023 10:48:06 -0700


myoung34 ***@***.***> wrote:

> Its worth noting you can (and should) be using TARGETPLATFORM as a
> build param, but my suggestions solve it regardless
> 

Interesting. So if I do a 

docker run --env "TARGETPLATFORM=linux/arm64"

then there is no need for you to change anything?

koen

myoung34:
> The GH_VERSION_NUMBER still appears:

That's just the diff output. Yes it will show context around changes, but that's not a change. Its simply visual

You can avoid that by using `-U` in git such as `git diff -U0 HEAD^ HEAD`

Per git docs:

```
   -U<n>, --unified=<n>
       Generate diffs with <n> lines of context instead of the usual three. Implies
       -p.
```

myoung34:
> docker run --env "TARGETPLATFORM=linux/arm64"

Note that Its an arg, not an env var. env vars are at runtime, args are only at build time.

`docker build --build-arg TARGETPLATFORM=linux/arm64` is how you'd set this because install_actions.sh is run at build time and not at run time there's no point in making it also an env var

koendv:
On Fri, 17 Mar 2023 11:09:06 -0700
myoung34 ***@***.***> wrote:

> > The GH_VERSION_NUMBER still appears:  
> 
> That's just the diff output. Yes it will show context around changes,
> but that's not a change. Its simply visual
> 
> You can avoid that by using `-U` in git such as `git diff -U0 HEAD^
> HEAD`
> 
> Per git docs:
> 
> ```
>    -U<n>, --unified=<n>
>        Generate diffs with <n> lines of context instead of the usual
> three. Implies -p.
> ```
> 

ok.

myoung34 (approved):

PR #288, opened 2023-03-22
Optimization

derskythe:
Hello, I optimized a couple of things for myself and found out that the finished image is reduced by almost 100 MB.
I combined the jobs into one single matrix. The new matrix does not require additional runs in case of latest/focal.
I also found that you are using old version of action scripts and some parts are hard coded.

*Changes:*
* combine all separate steps workflows to single matrix
* because latest is ubuntu-focal, remove this extra job
* added docker/metadata-action to clear tags generation
* update version of all workflow actions
* remove changes and rewrites for Dockerfiles and pass all necessary params via ARG
* provenance, sbom, BUILDTIME, VERSION, REVISION to docker/build-push-action
* more secure process of github.REF variable
* update action release to current fork
* in Dockerfile.base disable output on apt-get update and curl and remove all extra obtains of variables
* in Dockerfile all scripts now in RUNNER_DIR, remove unnecessary and fix calls

*BREAKING CHANGES:*
Run run workflows we need to setup repo variables:

1. BASE_IMAGE=owner/name-of-base
2. CHOWN_USER=runner
3. FORCE_CREATE_RELEASE=0
4. GH_RUNNER_VERSION=2.303.0
5. RELEASE_IMAGE_NAME=github-runner

You can see the results here:

https://github.com/derskythe/docker-github-actions-runner/actions/
https://hub.docker.com/repository/docker/derskythe/github-runner-base
https://hub.docker.com/repository/docker/derskythe/github-runner

### DIFF
--- a/.github/workflows/base.yml
+++ b/.github/workflows/base.yml
@@ -104,8 +104,6 @@ jobs:
           BUILDTIME=${{ fromJSON(steps.meta.outputs.json).labels['org.opencontainers.image.created'] }}
           VERSION=${{ fromJSON(steps.meta.outputs.json).labels['org.opencontainers.image.version'] }}
           REVISION=${{ fromJSON(steps.meta.outputs.json).labels['org.opencontainers.image.revision'] }}
-        secrets: |
-          GIT_AUTH_TOKEN=${{ secrets.GIT_AUTH_TOKEN }}
         platforms: linux/amd64,linux/arm64
         cache-from: type=gha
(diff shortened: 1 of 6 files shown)


myoung34:
Im not going to lie, this PR is **way** too big for me to know what youre trying to optimize.

Can you break it down into multiple PR's? I don't understand or agree with a lot of the changes youre proposing, and I mean that in the nicest way possible.

PR #289, opened 2023-03-28
Update Docker Compose to v2

### LINKED ISSUE #222: Install `docker compose` plugin for ARM compatibility.
Now that "compose 2" is a native plugin for docker itself (e.g. `docker compose` is a native subcommand), can you please install it instead of (or in addition to) the 1.27 x86 only `docker-compose`?

barnesew:
This PR updates docker compose to v2 (see differences between v1 and v2 [here](https://stackoverflow.com/a/66516826)). Compose is now installed via the docker compose plugin rather than installing the docker-compose executable as seen [here](https://docs.docker.com/compose/install/linux/). This PR also indirectly fixes #222.

In addition, I added a compatibility file located at `/usr/local/bin/docker-compose` which will allow executing docker compose commands with `docker-compose` instead of the new syntax `docker compose` to prevent breaking anyone's current process.

This has been successfully built on a MacBook (ARM64) as well as a CentOS 9 server (AMD64). The following command should print out the docker compose version after building the base image: `docker run --rm -it myoung34/github-runner-base:latest docker compose version`

Please let me know if there is anything else I can help test or fix!

### DIFF
--- a/Dockerfile.base
+++ b/Dockerfile.base
@@ -4,7 +4,6 @@ LABEL maintainer="myoung34@my.apsu.edu"
 ARG DUMB_INIT_VERSION="1.2.2"
 ARG GIT_CORE_PPA_KEY="A1715D88E1DF1F24"
 
-ENV DOCKER_COMPOSE_VERSION="1.27.4"
 ENV GIT_LFS_VERSION="3.2.0"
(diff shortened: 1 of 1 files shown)


tfish1 (approved):

myoung34:
Thanks!

PR #292, opened 2023-04-13
Fix docker compose shebang

wico-silva:
Using `docker-compose`  results in an error since https://github.com/myoung34/docker-github-actions-runner/pull/289 was merged:
```txt
bash: /usr/local/bin/docker-compose: /usr/bin/sh: bad interpreter: No such file or directory
```

In my container `/usr/bin/sh` doesn't exist. Updating the shebang to `/bin/sh` fixed the issue.


myoung34:
Which arch and base image? i want to make sure its `/bin/sh` everywhere to double check

wico-silva:
Got the error on `myoung34/github-runner:ubuntu-bionic` and `amd64`

PR #293, opened 2023-04-30
remove warning about compose on ARM

### LINKED ISSUE #290: Now Docker already supported ARM based
Check issue in README.md: https://github.com/docker/compose/issues/6831
Announcement: https://www.docker.com/blog/getting-started-with-docker-for-arm-on-linux/

Please remove this:
image

viseshrp:
First off, amazing stuff. Works like a charm on my Raspberry Pi 4gb. I just created a simple compose yml and pasted in your example including the token from my repo's runner settings. It all just works.

Check out this action which actually ran using your image on my Pi. https://github.com/viseshrp/website/actions/runs/4846104297/jobs/8635430719

So, I think the warning about compose is no longer valid.

Basically fixes #290

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,11 +31,6 @@ Also, some GitHub Actions Workflow features, like [Job Services](https://docs.gi
 
 Currently runners [do not support containerd](https://github.com/actions/runner/issues/1265)
 
-### Docker-Compose on ARM ###
-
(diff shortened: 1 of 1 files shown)


myoung34:
Yup, that needs to go. thanks

PR #298, opened 2023-05-15
Start Docker service with env variable

LordChunk:
Allow users to start the Docker service using an environment variable. This is useful when using [sysbox](https://github.com/nestybox/sysbox) since setting up Docker in an action would fail.


myoung34 (approved):

andrewhibbert:
I tried this out and got `/etc/init.d/docker: 103: ulimit: error setting limit (Operation not permitted)`. I've logged a ticket with sysbox too https://github.com/nestybox/sysbox/issues/702

PR #307, opened 2023-06-22
Use node LTS when building runner image

Hathoute:
Current images use an old version of nodejs (version 10). 
Running some workflows on these images fail because it assumes that node version is at least 16 ([GraphQL for example](https://github.com/unknorg/workflow-preprocessor/actions/runs/5346240353/jobs/9692968387)).

This PR builds the runners using the LTS version of nodejs.

### DIFF
--- a/Dockerfile.base
+++ b/Dockerfile.base
@@ -13,8 +13,9 @@ ENV DEBIAN_FRONTEND=noninteractive
 # hadolint ignore=SC2086,DL3015,DL3008,DL3013,SC2015
 RUN echo en_US.UTF-8 UTF-8 >> /etc/locale.gen \
   && apt-get update \
-  && apt-get install -y --no-install-recommends gnupg \
+  && apt-get install -y --no-install-recommends gnupg curl \
   && apt-key adv --keyserver keyserver.ubuntu.com --recv-keys ${GIT_CORE_PPA_KEY} \
+  && curl -sL https://deb.nodesource.com/setup_18.x | sudo -E bash - \
   && apt-get update \
   && apt-get install -y --no-install-recommends \
     gnupg \


myoung34:
👋 I like the idea, but I have a few asks:

1) I don't allow first-hand curl pipe bash [for security reasons](https://sysdig.com/blog/friends-dont-let-friends-curl-bash/)
2) If you notice I dont install language versions outside of what's available in first-hand repositories. There are exceptions to this which are largely (squinting because github cli isnt, ill actually fix this) pinned such as git and dumb-init

Is it possible instead to use node 17 in your [codeql workflow?](https://github.com/unknorg/workflow-preprocessor/blob/main/.github/workflows/codeql-analysis.yml)

such:

```
name: "CodeQL"

on:
  push:
    branches: [ main ]
  pull_request:
    branches: [ main ]
  schedule:
    - cron: '31 7 * * 3'

jobs:
  analyze:
    name: Analyze
    runs-on: ubuntu-latest
    permissions:
      actions: read
      contents: read
      security-events: write

    strategy:
      fail-fast: false
      matrix:
        language: [ 'TypeScript' ]
      matrix:
        node-version: [17.x]
    steps:
    - name: Checkout repository
      uses: actions/checkout@v3
    - name: Use Node.js ${{ matrix.node-version }}
      uses: actions/setup-node@v1
      with:
        node-version: ${{ matrix.node-version }}
    - name: Initialize CodeQL
      uses: github/codeql-action/init@v2
      with:
        languages: ${{ matrix.language }}
        source-root: src
    - name: Autobuild
      uses: github/codeql-action/autobuild@v2
    - name: Perform CodeQL Analysis
      uses: github/codeql-action/analyze@v2
```

Hathoute:
I used the setup-node action for that workflow. 

I just thought that it would make sense to have the LTS version available in runners, but I do agree that its better to not include anything other than what is already available in first hand repo. After all, one can just add a layer on top of your image for extreme cases.

PR #315, opened 2023-08-25
Docker service dying fix

kmkacpermajor:
When running without sudo, docker service dies as soon as I try to build an image

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -170,7 +170,7 @@ fi
 # Start docker service if needed (e.g. for docker-in-docker)
 if [[ ${_START_DOCKER_SERVICE} == "true" ]]; then
   echo "Starting docker service"
-  service docker start
+  sudo service docker start
 fi
 
 # Container's command (CMD) execution as runner user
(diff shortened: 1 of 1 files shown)


entrypoint.sh
```diff
@@ -170,7 +170,7 @@ fi
 # Start docker service if needed (e.g. for docker-in-docker)
 if [[ ${_START_DOCKER_SERVICE} == "true" ]]; then
   echo "Starting docker service"
-  service docker start
+  sudo service docker start
```
myoung34, line 173:
```suggestion

```

myoung34, line 173: this container can be run as root, meaning `sudo` isnt necesary 100% of the time

Can you wrap it in `if [[ $(id -u) -eq 0 ]];` and `if [[ ${_RUN_AS_ROOT} == "true" ]]; then` logic for that edge case?

myoung34 (changes requested):

entrypoint.sh
```suggestion
  _PREFIX=""
  [[ ${_RUN_AS_ROOT} != "true" ]] && _PREFIX="sudo"
  ${_PREFIX} service docker start
```

myoung34 (changes requested):

PR #333, opened 2023-10-26
Jasoncubic/feature/no default labels

JasonCubic:
Adds the environmental variable NO_DEFAULT_LABELS to prevent adding default labels by adding the parameter `--no-default-labels` to the config.sh call when creating the runner.

For reference, this is a link to the `--no-default-labels` parameter: https://github.com/actions/runner/blob/main/src/Runner.Listener/Runner.cs#L687

Preventing default labels is necessary in environments with multiple teams with self-hosted runners to deter teams from accidentally using an unintended runner.  Some of these teams may not be aware there are other self-hosted runners in the pool.


myoung34:
Thanks!

JasonCubic:
That's great @myoung34 do you know how long until this will make it into release?  I would like to use it.  😊

myoung34:
Releases are automated from upstream, so when those are tagged this will be tagged
Otherwise it goes out in latest nightly

PR #376, opened 2024-08-02
Update token.sh with set flags

Mearman:
- `set -e` causes the script to exit immediately if a command exits with a non-zero status.
- `set -u` causes the script to exit immediately if an unset variable is used.
- `set -o pipefail` causes the script to exit immediately if a pipe command fails.

### DIFF
--- a/token.sh
+++ b/token.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -euo pipefail
 
 _GITHUB_HOST=${GITHUB_HOST:="github.com"}
 
--- a/token.sh
+++ b/token.sh
@@ -15,7 +15,8 @@ API_HEADER="Accept: application/vnd.github.${API_VERSION}+json"
 AUTH_HEADER="Authorization: token ${ACCESS_TOKEN}"
 CONTENT_LENGTH_HEADER="Content-Length: 0"
 
-case ${RUNNER_SCOPE} in
+
+case ${RUNNER_SCOPE:-} in
   org*)
     _FULL_URL="${URI}/orgs/${ORG_NAME}/actions/runners/registration-token"
     ;;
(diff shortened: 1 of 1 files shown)


token.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -euo pipefail
```
myoung34, line 2:
```suggestion
```

There are lots of ways to fail, not all of which need to propagate up. Id rather be specific on which ones cause startup to fail since the container could throttle as a result. If theres a fail case that needs to cause this, make it specific with error messages etc

token.sh
```diff
@@ -14,7 +15,8 @@ API_HEADER="Accept: application/vnd.github.${API_VERSION}+json"
 AUTH_HEADER="Authorization: token ${ACCESS_TOKEN}"
 CONTENT_LENGTH_HEADER="Content-Length: 0"
 
-case ${RUNNER_SCOPE} in
+
+case ${RUNNER_SCOPE:-} in
```
myoung34, line 19: If theres a reason for it to be empty (which I think this handles) lets be explicit since its a new case. I dont think its valid to be empty, so it would cause new behavior here to fall into the `*)`

token.sh
```diff
@@ -26,10 +28,12 @@ case ${RUNNER_SCOPE} in
   *)
     _PROTO="https://"
     # shellcheck disable=SC2116
-    _URL="$(echo "${REPO_URL/${_PROTO}/}")"
+    _URL="$(echo "${REPO_URL:-/${_PROTO}/}" | sed "s#${_PROTO}##")"
```
myoung34, line 31: What is this for?

token.sh
```diff
@@ -26,10 +28,12 @@ case ${RUNNER_SCOPE} in
...
     _PATH="$(echo "${_URL}" | grep / | cut -d/ -f2-)"
     _ACCOUNT="$(echo "${_PATH}" | cut -d/ -f1)"
     _REPO="$(echo "${_PATH}" | cut -d/ -f2)"
+    if [[ -z "${_ACCOUNT}" ]]; then echo "Failed to parse account"; exit 1; fi
```
myoung34, line 35: im not sure these can fail, given they come from the URL

myoung34 (changes requested):

Mearman:
Sorry. Meant to mark this as non-draft. Still tinkering with different scenarios.
Thank you for the review though!

PR #384, opened 2024-08-31
feat: Prevent exposure of configuration environment variables to runner workflows

### LINKED ISSUE #383: Environment Variables for Runner Configuration Leaking to Workflows
As documented in the [README.md](https://github.com/myoung34/docker-github-actions-runner?tab=readme-ov-file#environment-variables), runner configuration can be set using environment variables passed to the container. However, we've discovered that these configuration variables are unintentionally accessible within the workflows themselves.

Our organization encountered a problem where environment variables that should not be present in workflows (beyond the [default environment variables](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables)) were leaking into the workflow context, causing unintended behavior.

While we've implemented a workaround for this issue in our organization, we would greatly appreciate if this could be addressed upstream as well.

### Dockerfile

```Dockerfile
FROM ghcr.io/myoung34/docker-github-actions-runner:2.319.1-ubuntu-jammy

...

COPY run.sh /
RUN chmod +x /run.sh

COPY pre-start-service.sh /
RUN chmod +x /pre-start-service.sh

ENTRYPOINT ["/run.sh"]
CMD ["./bin/Runner.Listener", "run", "--startuptype", "service"]
```

### run.sh

```bash
# Execute entrypoint.sh without starting the service, allowing us to unset environment variables after configuration but before runner startup
/entrypoint.sh true

# Execute pre-start operations
source /pre-start-service.sh

# Start the runner service
# Assuming RUN_AS_ROOT=false: https://github.com/myoung34/docker-github-actions-runner/blob/eb29a1af8cef6e75e9ea673beb78b090c646e45c/entrypoint.sh#L208
exec /usr/sbin/gosu runner "$@"
```

### pre-start-service.sh

```bash
# Unset environment variables used for myoung34/docker-github-actions-runner configuration
# to prevent unintended access from workflows
# Note: This unset operation must occur before the runner service starts
unset EPHEMERAL
unset DISABLE_AUTO_UPDATE
unset RANDOM_RUNNER_SUFFIX
unset ENTERPRISE_NAME
unset RUNNER_SCOPE
unset LABELS
unset RUNNER_GROUP
unset RUNNER_NAME_PREFIX
unset RUN_AS_ROOT
unset RUNNER_WORKDIR
unset APP_LOGIN
unset RUNNER_ALLOW_RUNASROOT
# Additional variables as needed...
```

## Ideal Solution

We propose implementing a mechanism to unset the configuration environment variables after runner setup (e.g., execution of config.sh) but before the runner is started (`./bin/Runner.Listener run --startuptype service`). This would ensure that these variables are available for configuration but not leaked to the workflow execution environment.

yamoyamoto:
closes #383 

1. Added a new environment variable flag `UNSET_CONFIG_VARS`. When set to `true`, it triggers the removal of all configuration environment variables after runner setup.
2. Refactored the usage of `CONFIGURED_ACTIONS_RUNNER_FILES_DIR` to use a local variable. This ensures proper functionality after unsetting environment variables.

## Testing Procedure

To verify the changes, follow these steps:

1. Build the Docker image:
   ```bash
   docker build -t yamoyamoto/github-runner .
   ```

2. Set up required environment variables:
   ```bash
   export REPO_URL='https://github.com/yamoyamoto/repo' 
   export RUNNER_NAME=runner-test
   export RUNNER_TOKEN=***** # Replace with actual token
   ```

3. Run the container without the new flag:
   ```bash
   docker run --rm --name github-runner \
     -e REPO_URL="${REPO_URL}" \
     -e RUNNER_NAME="${RUNNER_NAME}" \
     -e RUNNER_TOKEN="${RUNNER_TOKEN}" \
     yamoyamoto/github-runner
   ```

4. Run the container with the new flag enabled:
   ```bash
   docker run --rm --name github-runner \
     -e REPO_URL="${REPO_URL}" \
     -e RUNNER_NAME="${RUNNER_NAME}" \
     -e RUNNER_TOKEN="${RUNNER_TOKEN}" \
     -e UNSET_CONFIG_VARS="true" \
     yamoyamoto/github-runner
   ```

## Verification Results

To confirm the effectiveness of the changes, we used [hmarr/debug-action](https://github.com/hmarr/debug-action) to dump and inspect the environment variables in both scenarios.

### With UNSET_CONFIG_VARS=true

- The `REPO_URL` environment variable has been unset

image

### With default behavior (UNSET_CONFIG_VARS=false)

- The REPO_URL environment variable has not been unset

image

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -57,6 +57,7 @@ _RUNNER_GROUP=${RUNNER_GROUP:-Default}
 _GITHUB_HOST=${GITHUB_HOST:="github.com"}
 _RUN_AS_ROOT=${RUN_AS_ROOT:="true"}
 _START_DOCKER_SERVICE=${START_DOCKER_SERVICE:="false"}
+_UNSET_CONFIG_VARS=${UNSET_CONFIG_VARS:="false"}
 
 # ensure backwards compatibility
 if [[ -z ${RUNNER_SCOPE} ]]; then
@@ -151,6 +152,33 @@ configure_runner() {
 
 }
 
+unset_config_vars() {
+  echo "Unsetting configuration environment variables"
+  unset RUN_AS_ROOT
+  unset RUNNER_NAME
+  unset RUNNER_NAME_PREFIX
+  unset RANDOM_RUNNER_SUFFIX
+  unset ACCESS_TOKEN
+  unset APP_ID
+  unset APP_PRIVATE_KEY
+  unset APP_LOGIN
+  unset RUNNER_SCOPE
+  unset ORG_NAME
+  unset ENTERPRISE_NAME
+  unset LABELS
+  unset REPO_URL
+  unset RUNNER_TOKEN
+  unset RUNNER_WORKDIR
+  unset RUNNER_GROUP
+  unset GITHUB_HOST
+  unset DISABLE_AUTOMATIC_DEREGISTRATION
+  unset CONFIGURED_ACTIONS_RUNNER_FILES_DIR
+  unset EPHEMERAL
+  unset DISABLE_AUTO_UPDATE
+  unset START_DOCKER_SERVICE
+  unset NO_DEFAULT_LABELS
+  unset UNSET_CONFIG_VARS
+}
 
 # Opt into runner reusage because a value was given
 if [[ -n "${CONFIGURED_ACTIONS_RUNNER_FILES_DIR}" ]]; then
@@ -202,6 +230,11 @@ if [[ ${_START_DOCKER_SERVICE} == "true" ]]; then
   else
     ${_PREFIX} service docker start
   fi
+fi
+
+# Unset configuration environment variables if the flag is set
+if [[ ${_UNSET_CONFIG_VARS} == "true" ]]; then
+  unset_config_vars
 fi
 
 # Container's command (CMD) execution as runner user
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -58,6 +58,7 @@ _GITHUB_HOST=${GITHUB_HOST:="github.com"}
 _RUN_AS_ROOT=${RUN_AS_ROOT:="true"}
 _START_DOCKER_SERVICE=${START_DOCKER_SERVICE:="false"}
 _UNSET_CONFIG_VARS=${UNSET_CONFIG_VARS:="false"}
+_CONFIGURED_ACTIONS_RUNNER_FILES_DIR=${CONFIGURED_ACTIONS_RUNNER_FILES_DIR:-""}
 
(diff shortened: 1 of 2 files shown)


entrypoint.sh
```diff
@@ -189,6 +218,11 @@ if [[ ${_START_DOCKER_SERVICE} == "true" ]]; then
   ${_PREFIX} service docker start
 fi
 
+# Unset configuration environment variables if the flag is set
+if [[ ${_UNSET_CONFIG_VARS} == "true" ]]; then
+  unset_config_vars
```
myoung34, line 238: This is going to be problematic

The last section of this is 

```
# Container's command (CMD) execution as runner user


if [[ ${_RUN_AS_ROOT} == "true" ]]; then
  if [[ $(id -u) -eq 0 ]]; then
    "$@"
  else
    echo "ERROR: RUN_AS_ROOT env var is set to true but the user has been overridden and is not running as root, but UID '$(id -u)'"
    exit 1
  fi
else
  if [[ $(id -u) -eq 0 ]]; then
    [[ -n "${_CONFIGURED_ACTIONS_RUNNER_FILES_DIR}" ]] && chown -R runner "${_CONFIGURED_ACTIONS_RUNNER_FILES_DIR}"
    chown -R runner "${_RUNNER_WORKDIR}" /actions-runner
    # The toolcache is not recursively chowned to avoid recursing over prepulated tooling in derived docker images
    chown runner /opt/hostedtoolcache/
    /usr/sbin/gosu runner "$@"
  else
    "$@"
  fi
```

It still uses some of the ones you just uset here.

myoung34 (changes requested):

entrypoint.sh
yamoyamoto: Thank you for your thorough review!

> It still uses some of the ones you just unset here.

The variables used in the last section are only those prefixed with `_`, so I think there shouldn't be any issues with unsetting the environment variables without prefixes. What are your thoughts on this?

myoung34 (changes requested):
I used some free time to add basic tests to this project to make sure that all the logic is captured as assertions
If you'd like to squash all these commits down into one and rebase against master you'll see those run against this and it will help make sure that no user facing behavior is changed

myoung34 (approved):

myoung34:
Sweet. The tests make me more confident in these changes. If for some reason they didn't catch something I'll update the tests accordingly

myoung34:
It'll be out in latest, otherwise you'll have to wait for an upstream tag to be made from the official runners project. I don't retag or deviate from their tags by design

Issue #380, opened 2024-08-22
Question: where can I find the list of installed tools


DocMoebiuz:
For the official github runner images there is a page where they list all installed packages and tools:
https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2004-Readme.md

I was assuming that the tools would also be available for this container image version but I find that i.e. **AZ CLI** is missing.

Looks like the Github team has a json file which is considered for installing those tools:
https://github.com/actions/runner-images/blob/main/images/ubuntu/toolsets/toolset-2004.json

How did you go about these packages and tools? It looks like some of the packages are found in the Dockerfile.base

myoung34:
They're installed adhoc at best, but any improvement to that is definitely a consideration

vyadh:
I was considering doing a bit of refactoring of the Dockerfiles for similar reasons. What I'd actually like to do is turn things off to reduce the attack surface and reduce the need for patching.

Before I get carried away, how do you feel about me moving everything within the RUN in Dockerfiles.base into a dedicated script and then tidying it up a bit? The downside being that it would need to be copied into the image first, but it seems like a small price for easier maintainability? 

Once that's done, we could have a simple flags file of what to install, and the package list would be derived from that. Plus other useful configuration like user and group IDs.

If we did that, do we need two Dockerfiles or can we just have the one?

I'll note that while I can do testing on the things I use, it's not just, and I'm sure I might break things. Short of writing a test suite, how can we validate it?

myoung34:
I'm ok with a script and a manifest of sorts to run

It will still always be two dockerfiiles though, one for base and one to build off of when the version bumps for time to build reasons 

It also means people can provide their own base and lay the smaller dockerfile on it for a FROM

I'm also ok with a test to validate , it's been on my to-do for as long as Ive maintained this

DocMoebiuz:
Why do we have to recreate the wheel if the install scripts and config logic is already available https://github.com/actions/runner-images/blob/main/images/ubuntu/scripts/build/Install-Toolset.ps1

Ideally we could reuse those.

vyadh:
I was mainly doing just a bit of refactoring of what is there to make it easier for me to create a minimal image downstream as I'd rather use other container images for tool specific work. Using the VM tool installs sounds neat though, and presumably can be done without needing PowerShell installed looking at those install scripts. Not sure I'll be able to go that far, but maybe what I'm doing will make that slightly easier to accomplish.

[...]

PR #385, opened 2024-09-02
Refactored to a set of install scripts to avoid adding further complexity to the base Dockerfile and make it easier to disable tools

vyadh:
Refactored the `Dockerfile.base` out into a set of scripts that install the various tools in mostly the same way, just in a more maintainable and legible way. For my purposes the motivation was to make it easier to strip things out and create a minimal image. I've also changed a few things such as the removal of `lsb-release`, which seemed unnecessary, mainly due to it simplifying the refactoring and to avoid a Python dependency if its not otherwise required. I've tried to make the refactoring steps somewhat easier to follow / diffable via the various commits to make it easier to review.

It does a few less calls to `apt update` given the up-front sources configuration, and so is a touch faster, but it's not significant (a few 10s of seconds on my machine).

I've tested this and for my purposes it works, though I'm not using many of the tools that are installed in this image.

Somewhat related to #380.

### DIFF
--- a/build/install_base.sh
+++ b/build/install_base.sh
@@ -1,0 +1,74 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+function bootstrap_sources() {
+  apt-get install -y --no-install-recommends \
+      ca-certificates \
+      curl \
+      gnupg
+}
+
+function install_tools_apt() {
+  apt-get install -y --no-install-recommends \
+    tar \
+    unzip \
+    zip \
+    apt-transport-https \
+    sudo \
+    gpg-agent \
+    software-properties-common \
+    jq \
+    dirmngr \
+    locales \
+    dumb-init \
+    gosu \
+    build-essential \
+    zlib1g-dev \
+    zstd \
+    gettext \
+    libcurl4-openssl-dev \
+    inetutils-ping \
+    wget \
+    openssh-client \
+    python3-pip \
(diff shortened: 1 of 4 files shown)


build/install_base.sh
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env bash
+# hadolint ignore=SC2086,DL3015,DL3008,DL3013,SC2015
```
vyadh: I'm not sure what to do with this. I can change to shellcheck but it doesn't support the DL suppressions. I should probably figure out if they're even still needed but thought it would be worth your input first.

myoung34 (dismissed):
I used some free time to add basic tests to this project to make sure that all the logic is captured as assertions
If you'd like to squash all these commits down into one and rebase against master you'll see those run against this and it will help make sure that no user facing behavior is changed

vyadh:
Ah, great stuff. Thank you for doing that. I'll get on it.

build/install_base.sh
vyadh: I'm not getting any warnings, so I'm going to assume that these were all corrected during my refactoring.

vyadh:
I'll fix the shell check stuff next. I didn't notice there was an action for that. I was relying on IntelliJ telling me issues!

myoung34:
🤷 :shipit: 
Gotta trust in the tests lol

PR #393, opened 2024-09-09
Config driven install

vyadh:
This PR does a few things, mostly in pursuit of #380:
1) Installs all tools required by the runner itself (at least, as far as I can determine)
2) Installs all optional tools driven from a config file
a) apt installs are just a concatenated package list
b) script-based tools it takes the function name to install each tool from the config
3) Takes the uid and gid from a config file, which helps matching the runner user against host user configuration (since we have a build config file now anyway...)

I took the opportunity to split out `docker-cli` from `docker-ce`. The main reason being that if you're delegating docker builds to the host runner, you only need the cli. One extra `apt install` but I doubt this has much impact on the build time.


myoung34 (changes requested):
Revert the changes to README.md
I've captured them in a doc to put on the wiki instead of the README

myoung34 (approved):

myoung34:
Cool work, thanks! I'll update the wiki

vyadh:
Thanks for merging! This is going to be super useful to keep in sync with updates but retain a minimal image.

myoung34:
https://github.com/myoung34/docker-github-actions-runner/wiki/Usage#modifications

PR #397, opened 2024-09-23
Automatically Create Runner Groups via GitHub API

christopher-besch:
Add the `CREATE_RUNNER_GROUP` environment variable.

Instead of creating runner groups manually via the web interface, this variable creates them automatically on runner startup.

See: https://docs.github.com/en/enterprise-cloud@latest/rest/actions/self-hosted-runner-groups?apiVersion=2022-11-28#create-a-self-hosted-runner-group-for-an-organization

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,7 @@ RUN chmod +x /actions-runner/install_actions.sh \
   && rm /actions-runner/install_actions.sh \
   && chown runner /_work /actions-runner /opt/hostedtoolcache
 
-COPY token.sh entrypoint.sh app_token.sh /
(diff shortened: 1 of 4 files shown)


myoung34:
This is outside the scope of a runner itself. In the real world I would push to use something like a [controller](https://github.com/actions/actions-runner-controller/blob/90b68fec1a364e2eb1c784e190022c601bf23ecd/docs/managing-access-with-runner-groups.md) to manage this. The runners job is simply to come up and be available, not manage the existence of outside resources such as groups.

In addition this requires an enterprise level plan to manage groups so it's untestable on my side.

PR #403, opened 2024-10-05
Make sure there is no remnant runner config before configuring

### LINKED ISSUE #402: Ephemeral runner sometimes gets caught in a restart loop
An ephemeral runner can sometimes get caught in a restart loop. This happens if:

- The runner config is deleted in GitHub.
- The local runner config is still exists in /actions-runner/.runner

In this situation, https://github.com/myoung34/docker-github-actions-runner/blob/706339b472affde5736e39f7922494f271f9fd11/entrypoint.sh#L205 tries to configure the runner, but this fails due to the existing config: https://github.com/actions/runner/blob/ddf41af7678a8bc585d9e6b8ab70d941cdb687b2/src/Runner.Listener/Configuration/ConfigurationManager.cs#L109.

### Fix
We should delete any existing config in this situation, just to be safe.

### Logs

```
Runner reusage is disabled
Obtaining the token of the runner
Ephemeral option is enabled
Configuring

--------------------------------------------------------------------------------
|        ____ _ _   _   _       _          _        _   _                      |
|       / ___(_) |_| | | |_   _| |__      / \   ___| |_(_) ___  _ __  ___      |
|      | |  _| | __| |_| | | | | '_ \    / _ \ / __| __| |/ _ \| '_ \/ __|     |
|      | |_| | | |_|  _  | |_| | |_) |  / ___ \ (__| |_| | (_) | | | \__ \     |
|       \____|_|\__|_| |_|\__,_|_.__/  /_/   \_\___|\__|_|\___/|_| |_|___/     |
|                                                                              |
|                       Self-hosted runner registration                        |
|                                                                              |
--------------------------------------------------------------------------------
Cannot configure the runner because it is already configured. To reconfigure the runner, run 'config.cmd remove' or './config.sh remove' first.
Starting docker service
 * Starting Docker: docker
   ...done.

√ Connected to GitHub

Failed to create a session. The runner registration has been deleted from the server, please re-configure. Runner registrations are automatically deleted for runners that have not connected to the service recently.
```

### Analysis
I am not a 100% sure how this happens, but my theory is this:

- A workflow is cancelled manually in GitHub
- The runner is terminated
- It deregisters itself: https://github.com/actions/runner/blob/ddf41af7678a8bc585d9e6b8ab70d941cdb687b2/src/Runner.Listener/Configuration/ConfigurationManager.cs#L531
- Due to `stop_grace_period` being too short, it is terminated right afterwards.
- Thus it fails to remove the local config file: https://github.com/actions/runner/blob/ddf41af7678a8bc585d9e6b8ab70d941cdb687b2/src/Runner.Listener/Configuration/ConfigurationManager.cs#L542
- docker-compose automatically restarts the runner without recreating the container
- Thus the invalid local config file remains in /actions-runner/.runner

bodograumann:
Closes #402

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -26,6 +26,7 @@ deregister_runner() {
     RUNNER_TOKEN=$(echo "${_TOKEN}" | jq -r .token)
   fi
   ./config.sh remove --token "${RUNNER_TOKEN}"
+  [[ -f "/actions-runner/.runner" ]] && rm -f /actions-runner/.runner
   exit
 }
 
@@ -202,9 +203,7 @@ if [[ -n "${_CONFIGURED_ACTIONS_RUNNER_FILES_DIR}" ]]; then
 else
   echo "Runner reusage is disabled"
   if [[ ${_DEBUG_ONLY} == "false" ]]; then
-    # Make sure there is no remnant runner config, that would break registration
-    # Cf. https://github.com/myoung34/docker-github-actions-runner/issues/402
-    rm -f /actions-runner/.runner
+    [[ -f "/actions-runner/.runner" ]] && rm -f /actions-runner/.runner
     configure_runner
   fi
 fi
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -202,6 +202,9 @@ if [[ -n "${_CONFIGURED_ACTIONS_RUNNER_FILES_DIR}" ]]; then
 else
   echo "Runner reusage is disabled"
   if [[ ${_DEBUG_ONLY} == "false" ]]; then
+    # Make sure there is no remnant runner config, that would break registration
+    # Cf. https://github.com/myoung34/docker-github-actions-runner/issues/402
+    rm -f /actions-runner/.runner
     configure_runner
   fi
 fi


entrypoint.sh
```diff
@@ -202,6 +202,9 @@ if [[ -n "${_CONFIGURED_ACTIONS_RUNNER_FILES_DIR}" ]]; then
 else
   echo "Runner reusage is disabled"
   if [[ ${_DEBUG_ONLY} == "false" ]]; then
+    # Make sure there is no remnant runner config, that would break registration
```
```suggestion
```

entrypoint.sh
```diff
@@ -202,6 +202,9 @@ if [[ -n "${_CONFIGURED_ACTIONS_RUNNER_FILES_DIR}" ]]; then
...
+    # Cf. https://github.com/myoung34/docker-github-actions-runner/issues/402
```
```suggestion
```

no need for this breadcrumb, the commit will link to this PR

entrypoint.sh
```diff
@@ -202,6 +202,9 @@ if [[ -n "${_CONFIGURED_ACTIONS_RUNNER_FILES_DIR}" ]]; then
...
+    rm -f /actions-runner/.runner
```
```suggestion
[[ -f "/actions-runner/.runner" ]] && rm -f /actions-runner/.runner
```

myoung34 (changes requested):
Also please add `[[ -f "/actions-runner/.runner" ]] && rm -f /actions-runner/.runner` to `deregister_runner()` after `./config.sh remove ...`

bodograumann:
Thanks for the quick response @myoung34 . I have incorporated your suggestions.

myoung34 (approved):

PR #449, opened 2025-05-05
Remove `chmod +x` from Dockerfile

johankok:
Make the scripts executable in the repo and you can omit the `chmod +x` in the Dockerfile.


myoung34:
I appreciate the PR but I'd rather be explicit than implicit as new files may be added, its more obvious to see them need to be executable. The github UX for new files **and** permissions isnt obvious

PR #455, opened 2025-05-20
Fix RUNNER_WORKDIR not created if part of the folder structure is nonexisting

SaschaHenning:
If a RUNNER_WORKDIR is defined in the environment variables but the parent folder is not mapped to the container, the creation will fail.

**Sample:**
`RUNNER_WORKDIR: /tmp/runner/work`
results in console error
`mkdir: cannot create directory ‘/tmp/runner/work’: No such file or directory`

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -150,7 +150,7 @@ configure_runner() {
       --replace \
       "${ARGS[@]}"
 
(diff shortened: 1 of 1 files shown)


myoung34 (approved):

myoung34:
Thanks

PR #465, opened 2025-07-04
Allow hostname without github-runner prefix to avoid 64-char limit

ElsevierAlex:
Fix: Allow runner name to use hostname without the `github-runner-` prefix to avoid exceeding GitHub's 64-character limit for runner names.

Problem Description

Currently, the runner name generation logic in [[entrypoint.sh](https://github.com/myoung34/docker-github-actions-runner/blob/master/entrypoint.sh#L45)](https://github.com/myoung34/docker-github-actions-runner/blob/master/entrypoint.sh#L45) always includes a prefix when using the hostname, which can cause the total runner name to exceed GitHub's 64-character limit. 
The hostname length can be controlled by the consumer to be up to 64 characters, but the prefix may cause the limit to be exceeded, resulting in the runner failing to be created with the following error:
 `An error occurred: Not configured. Run config.(sh/cmd) to configure the runner.`
 
This is problematic in cases where:

- Adding the default prefix `github-runner-` (14 characters) + hostname would exceed the 64-character limit
- The current logic forces a dash separator even when `RUNNER_NAME_PREFIX` is empty


myoung34 (changes requested):
This could be a breaking change. 

Before: it would be `github-runner-{hostname}` with nothing set.
While your changes work, the after is now `hostname` with nothing set

If you Default `RUNNER_NAME_PREFIX` to `github-runner` I'll approve this as it means no one will need to change anything, and you can use `RUNNER_NAME_PREFIX=""` to get the new behavior for yourself

ElsevierAlex:
> This could be a breaking change.
> 
> Before: it would be `github-runner-{hostname}` with nothing set. While your changes work, the after is now `hostname` with nothing set
> 
> If you Default `RUNNER_NAME_PREFIX` to `github-runner` I'll approve this as it means no one will need to change anything, and you can use `RUNNER_NAME_PREFIX=""` to get the new behavior for yourself


> This could be a breaking change.
> 
> Before: it would be `github-runner-{hostname}` with nothing set. While your changes work, the after is now `hostname` with nothing set
> 
> If you Default `RUNNER_NAME_PREFIX` to `github-runner` I'll approve this as it means no one will need to change anything, and you can use `RUNNER_NAME_PREFIX=""` to get the new behavior for yourself

Addressed

myoung34 (approved):

Issue #369, opened 2024-06-10
Could not read private key from /dev/fd/63


Inveracity:
Heya, while attempting to register my runner with an app private key I ran into the following error message:

```sh
Could not read private key from /dev/fd/63
```

I found that I could work around it by making the following change in `app_token.sh`

```diff
+ PRIVATE_KEY_PATH=$1
rsa256_sign() {
+ openssl dgst -binary -sha256 -sign $PRIVATE_KEY_PATH
- openssl dgst -binary -sha256 -sign <(echo "$1")
}
```

and then run it

```sh
bash app_token.sh /path/to/my.pem
```

If this is a reasonable change I'll submit a PR

Host info:
- Ubuntu 24.04
- GNU Bash, version 5.2.21(1)-release (x86_64-pc-linux-gnu)
- OpenSSL 3.0.13 Jan 2024

myoung34:
Good catch, this was missed a long time ago. the current usage is definitely wrong as `<()` returns a file descriptor and not what was expected. A simple PR should be good enough

however scope the variable: 

```
rsa256_sign() {
 PRIVATE_KEY_PATH=$1
 openssl dgst -binary -sha256 -sign $PRIVATE_KEY_PATH
}
```
or 

```
rsa256_sign() {
 openssl dgst -binary -sha256 -sign $1
}
```

PR #498, opened 2025-09-26
fix app_token.sh

JontiH:
I found this from 
https://github.com/myoung34/docker-github-actions-runner/issues/369 

as I was having the same error `Could not read private key from /dev/fd/63` when using APP_PRIVATE_KEY
and after adding app_token.sh to my runner container with this fix, it worked

### DIFF
--- a/app_token.sh
+++ b/app_token.sh
@@ -61,7 +61,7 @@ base64url() {
 }
 
 rs256_sign() {
-    openssl dgst -binary -sha256 -sign <(echo "$1")
+    openssl dgst -binary -sha256 -sign echo "$1"
 }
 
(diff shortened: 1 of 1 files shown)


myoung34:
This is not right. `<(echo $1)` means to execute and return the contents
`openssl dgst -binary -sha256 -sign echo "$1"` means that it will run `openssl dgst -binary -sha256 -sign 'echo "$1"'` where -sign gets the literal `echo "{contents}"`

jharris-tc:
I just referred to your comment here https://github.com/myoung34/docker-github-actions-runner/issues/369#issuecomment-2158430298

PR #499, opened 2025-09-26
change APP_PRIVATE_KEY in readme to be more specific

JontiH:
I am not sure if I just did this wrong, or if it is in fact possible to pass the whole key as an env var, but I could only get this to work via a file and then passing the filepath.

Maybe this is stupid but having this readme fixed would have saved me some time for sure


myoung34:
It's not a path, it is in fact the contents: 

```
export _PRIVATE_KEY_CONTENTS=$(cat ~/Downloads/366b70228e3f46adab23cc16ce5a3cf5.2025-05-23.private-key.pem)
colima nerdctl -- run -it \
  -e RUNNER_NAME=testing-an-app \
  -e APP_ID=1313114 \
  -e APP_PRIVATE_KEY="${_PRIVATE_KEY_CONTENTS}" \
  -e APP_LOGIN=OctoKode \
  -e RUNNER_SCOPE=org \
  -e ORG_NAME=OctoKode \
  ghcr.io/myoung34/docker-github-actions-runner:latest
```

jharris-tc:
interesting, I wonder why I couldn’t get it to work.

could it be worth having a path variable as an alternative?

myoung34:
No, using sensitive files and paths to them are an anti pattern and creates a lot of exfil paths. Its generally a bad practice. Even most CSI like vault operator, etc still end up with contents as env vars for this reason. Using the contents is confirmed to work, so my suspicion is that the data is not correct. As long as its pem format, `export _PRIVATE_KEY_CONTENTS=$(cat ~/Downloads/366b70228e3f46adab23cc16ce5a3cf5.2025-05-23.private-key.pem)` works.

PR #523, opened 2025-11-24
Set fixed Docker GID to avoid collisions with some groups

ElsevierAlex:
Stabilise Docker GID to Prevent Conflicts With Base System Services

This PR resolves an issue where the Docker group GID changes unpredictably between image builds. This instability causes failures in Docker-related operations, especially when using the image in containerised GitHub Actions workflows (runs_on: container).

Across recent GitHub Actions Runner  image releases (e.g., 2.328.0 to 2.330.0), the GID assigned to the docker group has been inconsistent, landing somewhere in the 990–999 range.

Example 
```
GH_RUNNER_VERSION=2.328.0
polkitd:x:997:
docker:x:996:runner
runner:x:121:

and with:
GH_RUNNER_VERSION=2.329.0
docker:x:999:runner
runner:x:121:
```
 This GID drift breaks downstream images, the underlying cause is that several system services installed before Docker create groups that occupy the upper-range GIDs first. As a result, by the time the Docker package installs, the expected GID is already taken.

Common conflicting system groups include:
```
systemd-journal:x:999:
systemd-network:x:998:
polkitd:x:997:
```

To prevent future collisions, the PR pre-creates the docker group before any Docker-related installation occurs, assigning it a fixed, safe GID.

Docker group GID is now fixed at 500.
* The 500–700 range is widely considered a safe zone for non-core service accounts
* This ensures consistent behaviour across different base OS versions and avoids accidental overlap with systemd, polkit, or other system services.

Looking forward to hearing your feedback.

### DIFF
--- a/build/tools.sh
+++ b/build/tools.sh
@@ -38,25 +38,11 @@ function install_git-lfs() {
   rm -rf /tmp/lfs.tar.gz "/tmp/git-lfs-${GIT_LFS_VERSION}"
 }
 
-function configure_docker_group_id() {
-  local desired_gid="$(docker_group_id)"
-  local current_gid=$(getent group docker | cut -d: -f3)
-
-  if [[ "$current_gid" != "$desired_gid" ]]; then
-    # Expected to fail if the group already exists or the GID is already in use
-    groupadd --system --gid "$desired_gid" docker 2>/dev/null || true
-  fi
-}
-
 function install_docker-cli() {
-  configure_docker_group_id
-
   apt-get install -y docker-ce-cli --no-install-recommends --allow-unauthenticated
 }
 
 function install_docker() {
-  configure_docker_group_id
-
   apt-get install -y docker-ce docker-ce-cli docker-buildx-plugin containerd.io docker-compose-plugin --no-install-recommends --allow-unauthenticated
 
   echo -e '#!/bin/sh\ndocker compose --compatibility "$@"' > /usr/local/bin/docker-compose
--- a/build/tools.sh
+++ b/build/tools.sh
@@ -38,11 +38,25 @@ function install_git-lfs() {
   rm -rf /tmp/lfs.tar.gz "/tmp/git-lfs-${GIT_LFS_VERSION}"
 }
 
+function configure_docker_group_id() {
+  local desired_gid="$(docker_group_id)"
+  local current_gid=$(getent group docker | cut -d: -f3)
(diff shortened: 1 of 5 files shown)


build/tools.sh
```diff
@@ -38,11 +38,25 @@ function install_git-lfs() {
   rm -rf /tmp/lfs.tar.gz "/tmp/git-lfs-${GIT_LFS_VERSION}"
 }
 
+function configure_docker_group_id() {
```
myoung34: This should go around [here](https://github.com/myoung34/docker-github-actions-runner/blob/ddaa37c0295a92eb119932f5d708e54129d5d03b/build/install_base.sh#L57) instead, no need to make it a function, just do `groupadd -g "$(docker_group_id)" docker || :`

myoung34 (changes requested):

build/tools.sh
ElsevierAlex: Thanks for the suggestion!  Moving it to [here](https://github.com/myoung34/docker-github-actions-runner/blob/ddaa37c0295a92eb119932f5d708e54129d5d03b/build/install_base.sh#L57), unfortunately, doesn't work. The Docker packages automatically create the docker group with GID 999 during `install_tools_apt`. We need to create it with GID 500 beforehand to prevent this. 
```
getent group docker
docker:x:999:runner
groupadd -g 500  docker || :
groupadd: group 'docker' already exists
```

myoung34: You can add it after https://github.com/myoung34/docker-github-actions-runner/blob/ddaa37c0295a92eb119932f5d708e54129d5d03b/build/install_base.sh#L52 just before install_tools_apt and install_tools so that it exists before those run, just add a comment above such as `# The docker group needs to run before installers` or similar

ElsevierAlex: done

myoung34 (changes requested):
.

myoung34 (approved):

myoung34 (changes requested):
Last nit, if you'll modify https://github.com/myoung34/docker-github-actions-runner/blob/85021afac43456607cecd0366befbe7c9dc7fdb5/goss_base.yaml#L86 to have 

```
group:
    runner:
        exists: true
        gid: 121
    docker:
        exists: true
        gid: 500
```
then the tests will verify that the group change works as expected

Issue #278, opened 2023-02-01
Allow sourcing of env vars from file/secrets


mmastrac:
The current setup requires hardcoding access tokens in `docker-compose.yml` files (or the equivalent), but docker's secrets infrastructure is a little bit more secure overall.

Ideally an environment variable could be used, and if that variable is set, `entrypoint.sh` would source that file before running anything else. The user could then create a local secret file, and specify it like so:

`ENV_FILE=/var/run/secrets/my-secrets`

The secrets file would be a set of environment key/value pairs, though it would likely just be an `ACCESS_TOKEN` or `RUNNER_TOKEN` entry:

`ACCESS_TOKEN=xyz`

myoung34:
Can you not use the `.env` file method documented [here](https://docs.docker.com/compose/environment-variables/set-environment-variables/) ?

or `env_file: ` [here](https://docs.docker.com/compose/environment-variables/set-environment-variables/#use-the-env_file-attribute)

mmastrac:
That works! I would prefer something that explicitly supports the secrets infrastructure, but it's a good enough workaround for what I need. Thanks for the heads-up.

myoung34:
Environment variables are never secret, but alternatively id rather keep this container unaware of the orchestration nuances

PR #528, opened 2025-11-27
Adding Docker Swarm secrets support

sweco-semhul:
Adding support for Docker Swarm secrets as option for sensitive variables.

And as a better solution for this issue, when running in a Docker Swarm environment, https://github.com/myoung34/docker-github-actions-runner/issues/278

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -4,6 +4,21 @@
 export RUNNER_ALLOW_RUNASROOT=1
 export PATH=${PATH}:/actions-runner
 
+# Function to read secrets from file in a Docker Swarm setup
+read_secret() {
+  local secret_name="$1"
+  local secret_file="/run/secrets/${secret_name}"
+  if [ -f "${secret_file}" ]; then
+    export "${secret_name}"="$(cat ${secret_file})"
+  fi
+}
+
+# Read Docker secrets if available
+read_secret "ACCESS_TOKEN"
+read_secret "RUNNER_TOKEN"
+read_secret "APP_ID"
+read_secret "APP_PRIVATE_KEY"
+
 # Un-export these, so that they must be passed explicitly to the environment of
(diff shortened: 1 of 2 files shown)


entrypoint.sh
```diff
@@ -4,6 +4,21 @@
 export RUNNER_ALLOW_RUNASROOT=1
 export PATH=${PATH}:/actions-runner
 
+# Function to read secrets from file in a Docker Swarm setup
```
myoung34, line 7: Im ok with this if its in a new file `secrets.sh`

```
#!/usr/bin/env bash
set -euo pipefail

SECRETS_DIR="/run/secrets"

for secret_file in "$SECRETS_DIR"/*; do
  # Skip if no files matched
  [[ -e "$secret_file" ]] || continue

  secret_name=$(basename "$secret_file")
  env_var_name=$(echo "$secret_name" | tr '[:lower:]' '[:upper:]' | tr '-' '_')

  # Read the secret
  secret_value=$(cat "$secret_file")

  # Export it
  export "$env_var_name=$secret_value"

  echo "Loaded secret: $env_var_name"
done

echo "Docker Swarm secrets loaded."
```

Then in entrypoint.sh at the top: `_DOCKER_SWARM_SECRETS_ENABLED=${DOCKER_SWARM_SECRETS_ENABLED:-false}`

and somewhere furhter below:

```
if [[ "${_DOCKER_SWARM_SECRETS_ENABLED}" == "true" ]]; then
  echo "Docker Swarm secrets enabled — loading secrets..."
  /secrets.sh
fi
```

also dont forget to modify https://github.com/myoung34/docker-github-actions-runner/blob/master/Dockerfile#L22-L23 to add script.sh with +x but also mkdir -p /run/secrets as well somewhere

myoung34 (changes requested):

PR #570, opened 2026-03-12
[WIP] bash cleanup + minor changes

laur89:
Mostly bash cleanup/simplification. Individual commit messages have needed/relevant info.

Few noteworthy _functional_ changes:

- note until now `deregister_runner()` was broken if `UNSET_CONFIG_VARS` was set, as `deregister_runner()` references some config vals that would've been unset beforehand.
  - I'd consider nuking `UNSET_CONFIG_VARS` altogether and remove that logic; or instead of `unset`ting, de-export all of them, like we're already doing for a select variables at the top of `entrypoint.sh`?
- curl & jq invocations get added -f/-e flags to fail fast;
- in app_token/request_access_token() we now assert the chosen app really does have required permissions set. this will help users when setting up the App & onboarding docker-github-runner itself.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,7 @@ RUN chmod +x /actions-runner/install_actions.sh \
   && rm /actions-runner/install_actions.sh \
   && chown runner /_work /actions-runner /opt/hostedtoolcache
 
-COPY token.sh entrypoint.sh app_token.sh common.sh /
+COPY          token.sh  entrypoint.sh  app_token.sh common.sh /
 RUN chmod +x /token.sh /entrypoint.sh /app_token.sh
 
 ENTRYPOINT ["/entrypoint.sh"]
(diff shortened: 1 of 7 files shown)


myoung34:
There are too many changes (and some unrelated to each other) in this PR for me to review
Please split this up into smaller PRs, but avoid making changes that do not add or remove functionality **especially** ones that are breaking changes (changing var names that users are setting). A lot of this does not objectively improve this project and are not likely going to be approved such as

* removing/changing the github host, etc
* setting flags to fail fast. these dont fail often (ever) and when they do its only in build so theres no benefit
* removing things like | tr. this is hard to test for side effects and again: no benefit
* logging isnt currently an issue and just adds more code to call around. `|| echo foo` isnt much better than `|| fail` 

some might be, but i cant in good faith let this PR go through with so much change going on

laur89:
> removing/changing the github host, etc

Not the case

> setting flags to fail fast. these dont fail often (ever) 

It failed for me with no indication that a specific app permission (and which one, who knows right) was missing. But if guesswork is the point, then I stand corrected.

> removing things like | tr

Simple to test against side effects, but fair.

> logging isnt currently an issue

Intention wasn't logging, but more concise way of bailing, i.e. `method "fail with this message"` as opposed to `( echo "why is subshell here even needed for bailing?"; exit 1 )`

---

But to each their own 🤷🏼‍♂️

PR #582, opened 2026-04-08
Trap EXIT to deregister runner when runner exits by itself

### LINKED ISSUE #581: Deregistration doesn't happen when runner process exits or crashes
Currently, if the runner application exits either normally (e.g. runner auto updates) or due to an error (e.g., exit code 1), the runner is not deregistered automatically.

In my docker compose file, I have set `restart: always` to try to keep runners available, but if the runner process exits and then the container is restarted, a new runner instance is registered and the old instance is orphaned.  Our runner was continuously exiting due to errors at launch, and we ended up registering 10,000 runner instances, hitting the limit of allowed runner instances.

From what I can tell, this seems to happen because the [deregister trap set up in `entrypoint.sh`](https://github.com/myoung34/docker-github-actions-runner/blob/e323dd3fa305995209c4a2b48c9b575087a6547f/entrypoint.sh#L239) only catches external signals to kill the runner like `SIGTERM` or `SIGINT`, but the runner exiting by itself isn't covered.

I think if we add EXIT to the list of signals to catch, that should allow deregistration to happen when the runner process exits by itself:

```bash
trap_with_arg deregister_runner SIGINT SIGQUIT SIGTERM INT TERM QUIT EXIT
```

jonathan528i:
Currently, if the runner application crashes or exits natively with an error (e.g., exit code `1`), the `deregister_runner` function in `entrypoint.sh` is never called. This leaves orphaned runner instances in github.  Closes #581.

The current trap code only looks for external signals to kill/stop the runner process, but not when the runner process exits by itself.  By also trapping `EXIT`, we can ensure `deregister_runner` is called whenever the runner exits.

### Changes

* `entrypoint.sh`: Added `EXIT` to the `trap_with_arg` list so the script guarantees deregistration regardless of how the container's main process terminates.
* `goss_trap_exit.yaml`: Added a new test to validate if exit code 1 is received, that deregister happens correctly.
* `.github/workflows/test.yml`: Integrated the new test into the CI pipeline for both Ubuntu and Debian builds.

### Testing

Locally ran the new test without the change to trap and verified it failed.  Re-ran the test with the change and verified it passes.  Also ran `goss_full.yaml` and `goss_reusage_fail` to ensure the change didn't break existing tests.


myoung34:
The new test looks like theres a race condition to pass

jonathan528i:
> The new test looks like theres a race condition to pass

My best guess is due to the `exit` call in `deregister_runner` function.  I'm not sure if the `exit` call in different bash environments will either return the error code that triggered the trap, or the last executed command before the `exit` call in the function (which would be 0, therefore causing the test to fail since it's expecting 1)?

I've updated the code but we'll have to see if that fixes the tests once the pipeline runs again.

myoung34:
Seems to always fail for arm64

myoung34:
My concern is how flakey these tests are; arm64 fails consistently

entrypoint.sh
```diff
@@ -20,6 +20,7 @@ trap_with_arg() {
 }
 
 deregister_runner() {
+  COMMAND_EXIT_CODE=$?
```
myoung34, line 23: the issue here is 2 things:

* that [EXIT doesnt trap signals](https://unix.stackexchange.com/questions/520035/exit-trap-with-posix?utm_source=chatgpt.com)
* qemu is doing something differently with how EXIT works here (why arm64 only, cant confirm _what_ though)

```suggestion
  local COMMAND_EXIT_CODE="${1:-EXIT}"
  echo "Caught ${COMMAND_EXIT_CODE} - Deregistering runner"
```

myoung34 (changes requested):

PR #593, opened 2026-05-10
feat(entrypoint): add multi-repo mode via REPO_URLS

opastorello:
## Summary

Adds a multi-repo mode to `entrypoint.sh` so a single container can run N self-hosted runners, one per repo, via:

```
-e REPO_URLS=https://github.com/me/repo1,https://github.com/me/repo2,https://github.com/me/repo3
```

This is useful for accounts (user or org) where you want a single container handling several repos without standing up one container per repo.

When `REPO_URLS` is empty, the existing single-repo flow runs unchanged — full backwards compat.

## Behavior

- **Auth minted once.** `ACCESS_TOKEN` (PAT) or `APP_*` produces an access token reused across all repos. Per-repo registration tokens are obtained via N separate `token.sh` calls at configure time.
- **Per-runner state** lives in `/actions-runner-N`, hardlinked from `/actions-runner` (`cp -al`) so `bin`/`externals` share inodes — disk cost is per-runner-state-only, not N × ~380 MB. Falls back to `cp -r` if cross-FS.
- **Per-runner naming:** `\${RUNNER_NAME_PREFIX:-github-runner}-<repo_short>-<suffix>` with workdir under `\${RUNNER_WORKDIR:-/_work}/<runner_name>`.
- **Signal trap** (`deregister_all`) iterates all configured runners, refreshes the App token if applicable, mints a fresh per-repo registration token, and `./config.sh remove`s each.
- **Fork & wait:** N background `Runner.Listener run` processes, then `wait`. Container's `CMD` is ignored in multi-mode.

## Constraints in multi-mode

- `RUNNER_SCOPE` is forced to `repo`
- `CONFIGURED_ACTIONS_RUNNER_FILES_DIR` (reusage) is rejected — hard error
- `RUNNER_TOKEN` with >1 URL is rejected (registration tokens are per-repo)

## Tests

- New `goss_multi_repo.yaml` asserts the multi-mode `DEBUG_ONLY` output (Preparing lines, multi-mode header, per-runner sub-blocks for 2 repos)
- Wired into both `ubuntu_tests` and `debian_tests` matrices in `test.yml`
- `README.md` updated with the new `REPO_URLS` env var

## Test plan

- [x] Local sandbox dry-run of `entrypoint.sh` multi-mode produces output matching every line of `goss_multi_repo.yaml`
- [x] Real container (`docker build` from this repo + `docker run` with `REPO_URLS=...,DEBUG_ONLY=true`) produces identical output
- [x] Single-mode (`REPO_URL=...`) produces identical output to pre-change behavior
- [x] `REPO_URLS` + `CONFIGURED_ACTIONS_RUNNER_FILES_DIR` exits 1 with the expected error
- [x] `RUNNER_TOKEN` with 2 URLs exits 1 with the expected error
- [ ] CI runs all 5 goss specs across ubuntu-{focal,jammy,noble} and debian-{bookworm,trixie} on amd64+arm64


myoung34:
I dont think Im willing to support this

If the runner supported it, this would but this is too much complexity for that

I personally do not believe in having PID tracking in a container as it complicates error handling, signaling and even user behavior

Im sure this will be some philosophical disagreement, but this project will only support multiple repos if the underlying agent does in its core functionality

opastorello:
Thanks for taking the time to look at this and for being upfront about the philosophical fit — that's fair, and it's your call to make.

The PID tracking concern is reasonable. I'd rather respect the project's direction than push a pattern you'd have to maintain reluctantly.

Two quick questions before I close this out, if you don't mind:

1. For the underlying need (one host serving N repos without standing up N containers), is there a pattern you'd recommend pointing users at? Docker Compose with N services sharing a volume, a k8s `Deployment` with `replicas`, something else? Happy to send a docs-only PR if there's a canonical answer worth surfacing in the README.
2. If/when `actions/runner` ever grows native multi-repo support, would you want a heads-up issue tracking that upstream, or do you prefer to pick it up organically?

Either way, no hard feelings — I'll plan to close this PR once you've had a chance to weigh in. Thanks again.

myoung34:
1) yep, just whatever is native to the orchestrator such as a k8s deployment with replicas, nomad groups?, etc
2) I dont really track upstream changes but I'm happy to leave an issue here that tracks an upstream issue

PR #610, opened 2026-06-10
fix: deregister runner on exit and preserve exit code

ElsevierAlex:
Issue: The bare exit at [entrypoint.sh](https://github.com/myoung34/docker-github-actions-runner/blob/master/entrypoint.sh#L54) returns the exit code of `config.sh remove` rather than the original process exit code.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -29,7 +29,8 @@ deregister_runner() {
   fi
   _DEREGISTERED=true
 
-  echo "Caught ${1:-EXIT} - Deregistering runner"
+  local CAUGHT="${1:-EXIT}"
+  echo "Caught ${CAUGHT} - Deregistering runner"
 
   if [[ -n "${ACCESS_TOKEN}" ]]; then
     # If using GitHub App authentication, refresh the access token before deregistration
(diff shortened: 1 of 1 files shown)


entrypoint.sh
```diff
@@ -22,13 +22,14 @@ trap_with_arg() {
...
+  local _EXIT_CODE=$?
+
   if [[ "${_DEREGISTERED}" == "true" ]]; then
     return
   fi
   _DEREGISTERED=true
 
-  local CAUGHT="${1:-EXIT}"
-  echo "Caught ${CAUGHT} - Deregistering runner"
+  echo "Caught ${1:-EXIT} - Deregistering runner"
```
myoung34: we need to keep EXIT into _EXIT_CODE since EXIT doesnt return a value. see https://github.com/myoung34/docker-github-actions-runner/pull/596

myoung34 (changes requested):

entrypoint.sh
ElsevierAlex: done

myoung34:
I threw AI at it to get a little closer to what should resolve this instead of back and forth, but this should be resolved in #611 

Thanks @ElsevierAlex

PR #613, opened 2026-06-23
fix: respect EPHEMERAL=false and EPHEMERAL=0 to disable ephemeral mode

CSchulz:
## Problem

The `EPHEMERAL` check in `entrypoint.sh` uses `[ -n "${EPHEMERAL}" ]`, which tests whether the variable is **non-empty**. This means any non-empty string — including `false` and `0` — enables the `--ephemeral` flag:

```bash
EPHEMERAL=false  # non-empty → --ephemeral is added  ❌
EPHEMERAL=0      # non-empty → --ephemeral is added  ❌
EPHEMERAL=true   # non-empty → --ephemeral is added  ✓
```

It is currently impossible to explicitly disable ephemeral mode by setting the environment variable to a falsy value.

## Fix

Add explicit `!= "false"` and `!= "0"` guards to the condition:

```bash
# Before
if [ -n "${EPHEMERAL}" ]; then

# After
if [ -n "${EPHEMERAL}" ] && [ "${EPHEMERAL}" != "false" ] && [ "${EPHEMERAL}" != "0" ]; then
```

## Behavior after fix

| `EPHEMERAL` value | Before | After |
|---|---|---|
| unset | not ephemeral | not ephemeral ✓ |
| `true` | ephemeral | ephemeral ✓ |
| `1` | ephemeral | ephemeral ✓ |
| `false` | ephemeral ❌ | not ephemeral ✓ |
| `0` | ephemeral ❌ | not ephemeral ✓ |

This is a **non-breaking change** — all existing behavior for unset, `true`, and `1` is preserved.

## Context

Users running persistent (non-ephemeral) runners in private repositories need to be able to set `EPHEMERAL=false` to prevent the container from restarting after every job. The current behavior makes this impossible.

### DIFF
--- a/.github/workflows/build-fork.yml
+++ b/.github/workflows/build-fork.yml
@@ -1,0 +1,48 @@
+name: Build fork image (ubuntu-noble, amd64)
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - master
+    paths:
+      - 'entrypoint.sh'
+      - 'Dockerfile'
+
+permissions:
+  contents: read
+  packages: write
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@d7f5e7f509e45cec5c76c4d5afdd7de93d0b3df5 # v4.1.0
+
+      - name: Log in to GitHub Container Registry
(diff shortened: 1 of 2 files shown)


myoung34:
This is by design and always has been. The presence of it in any capacity enables it

CSchulz:
> This is by design and always has been. The presence of it in any capacity enables it

Thanks for you reply, what do you mean by "The presence of it in any capacity enables it"?

myoung34:
Providing anything other than an empty string adds `--ephemeral` to the runner args

PR #621, opened 2026-07-06
feat: add RUNNER_WORKDIR_ROOT to override the default workdir base

jonbackhaus:
## What

Adds an optional **`RUNNER_WORKDIR_ROOT`** environment variable (default: `/_work`) that overrides the base directory used to build the default per-runner workdir.

When `RUNNER_WORKDIR` is **not** set, the workdir resolves to `<RUNNER_WORKDIR_ROOT>/<runner name>`. Since the runner name is already unique per runner (random suffix or prefix-based), replicas that share a mount each get their own subdirectory without colliding. An explicit `RUNNER_WORKDIR` still takes precedence and is used as-is.

## Why

Runners that want to share a common root (e.g. a shared build-cache bind mount) previously had no way to point the default `/_work` base elsewhere while keeping the per-runner isolation. Setting `RUNNER_WORKDIR_ROOT=/some/root/path` lets replicas auto-deploy with each runner getting `/some/root/path/<random name>`.

### Note on PR

This supersedes the earlier `ENABLE_RUNNER_WORKDIR_SUFFIX` + cleanup-on-exit approach from this PR, per review feedback: the cleanup had too many side effects, and the suffixing is already handled by the existing runner-name default — so all that's needed is an overridable root.

## Workdir resolution

| `RUNNER_WORKDIR` | `RUNNER_WORKDIR_ROOT` | Resulting workdir |
|---|---|---|
| unset     | unset               | `/_work/<name>` (historical default) |
| unset     | `/shared`           | `/shared/<name>` |
| `/explicit` | unset             | `/explicit` |
| `/explicit` | `/shared`         | `/explicit` (explicit wins) |

## Changes

- `entrypoint.sh`: single-line workdir default now honors `RUNNER_WORKDIR_ROOT`; removed the `ENABLE_RUNNER_WORKDIR_SUFFIX` flag, the suffix-append block, and `cleanup_runner_workdir()`; debug output now prints the resolved root.
- `README.md`: documented `RUNNER_WORKDIR_ROOT`.
- `goss_workdir_root.yaml` + `.github/workflows/test.yml`: goss test (wired into both matrix jobs) asserting `RUNNER_WORKDIR_ROOT` overrides the default base.

## Testing

- Debug runs verified across all four cases above, including `RUNNER_WORKDIR_ROOT=/mnt/tank00/dataset00/path` → `.../<name>`.
- `shellcheck entrypoint.sh`: clean.
- Added goss coverage for the root-override path.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


entrypoint.sh
```diff
@@ -21,6 +21,19 @@ trap_with_arg() {
...
+cleanup_runner_workdir() {
+  # Only owns/cleans a unique per-runner path when the suffix feature is on;
+  # otherwise the workdir may be shared or externally managed and must be
+  # left untouched.
+  if [[ "${_ENABLE_RUNNER_WORKDIR_SUFFIX}" != "true" ]]; then
+    return
+  fi
+  if [[ -n "${_RUNNER_WORKDIR}" ]] && [[ -d "${_RUNNER_WORKDIR}" ]]; then
+    echo "Cleaning up runner workdir ${_RUNNER_WORKDIR}"
+    rm -rf "${_RUNNER_WORKDIR}" || echo "WARNING: Failed to clean up runner workdir ${_RUNNER_WORKDIR}"
```
myoung34: I've intentionally avoided this as this can take a lot of time and when this is not a persistent store will be cleaned up by the orchestration

myoung34 (changes requested):

entrypoint.sh
jonbackhaus: If you like the `ENABLE_RUNNER_WORKDIR_SUFFIX` behavior otherwise, I can either gate this behind a flag or remove it entirely; your call.

entrypoint.sh
```diff
@@ -81,7 +96,18 @@ if [[ ${RANDOM_RUNNER_SUFFIX} != "true" ]]; then
   fi
 fi
 
-_RUNNER_WORKDIR=${RUNNER_WORKDIR:-/_work/${_RUNNER_NAME}}
```
myoung34, line 84: If you pass -e `RUNNER_WORKDIR=""` you can make this be `/_work/${_RUNNER_NAME}` which is unique per runner

jonbackhaus: Agreed, but I don't think that will work on my system? (I'm running TrueNAS, so my bind mount path looks like `/mnt/tank00/dataset00/path`.) I was hoping to enable replicas to auto-deploy multiple runners. So the root build path would be common but each runner would get a sub-directory based on the random name.

myoung34: `-e RUNNER_WORKDIR="" -e RUNNER_NAME=foo-$(head /dev/urandom | tr -dc A-Za-z0-9 | head -c 13 ; echo '')`

myoung34: I'd be willing to support a `RUNNER_WORKDIR_ROOT` that can be passed in or default to `/_work` in lieu of this PR

entrypoint.sh
myoung34: Lets remove the clean up entirely, it will have many side effects even when enabled

entrypoint.sh
```diff
@@ -81,7 +96,18 @@ if [[ ${RANDOM_RUNNER_SUFFIX} != "true" ]]; then
...
+# Base workdir defaults to /_work; an explicit RUNNER_WORKDIR overrides it.
+_RUNNER_WORKDIR=${RUNNER_WORKDIR:-/_work}
+_ENABLE_RUNNER_WORKDIR_SUFFIX=${ENABLE_RUNNER_WORKDIR_SUFFIX:-false}
```
myoung34: lets keep `RUNNER_WORKDIR_ROOT` to override `/_work` and remove everything else since it already has support for random suffix by using the runner name

myoung34 (changes requested):

entrypoint.sh
jonbackhaus: I like that implementation; let me get on it and update the PR. Thanks @myoung34 !

jonbackhaus (commented):
Changes implemented as-requested.

myoung34:
Tests are broken due to some upstream goss stuff, but this is fine

